PR #23, opened 2021-03-05
Added Host deconstructor

acameron13:


anyaevostinar:
Looks great, could you add deletion of the repro_syms as well to be thorough?

PR #45, opened 2021-07-02
Lysogeny dev

acameron13:
Added basic implementation of phage lysogeny with corresponding tests, with some improvements to modularity of code.


SymSettings.cfg
```diff
@@ -1,32 +1,35 @@
-### DEFAULT ###
```
anyaevostinar, line 1: Please make sure that the SymSettings stay the default

anyaevostinar (approved):

PR #58, opened 2021-07-19
Zhen sgg host and syms

renztim:
Sgg host and sym

### DIFF
--- a/SymSettings.cfg
+++ b/SymSettings.cfg
@@ -1,42 +1,46 @@
 ### MAIN ###
 # Global Settings
 
-set SEED 10                    # What value should the random seed be? If seed <= 0, then it is randomly re-chosen.
-set COMPETITION_MODE 0         # Should a competition between two types be conducted? (Which is specified in the code)
-set DATA_INT 100               # How frequently, in updates, should data print?
-set MUTATION_SIZE 0.002        # Standard deviation of the distribution to mutate by
-set HORIZ_MUTATION_SIZE -1     # Standard deviation of the distribution to mutate by for horizontal transmission, if -1 MUTATION_SIZE used
-set MUTATION_RATE 1            # Value 0 to 1 of probability of mutation
-set HORIZ_MUTATION_RATE -1     # Value 0 to 1 of probability of mutation for horizontal transmission, if -1 MUTATION_RATE used
-set SYNERGY 5                  # Amount symbiont's returned resources should be multiplied by
-set VERTICAL_TRANSMISSION 0.7  # Value 0 to 1 of probability of symbiont vertically transmitting when host reproduces
-set HOST_INT -2                # Interaction value from -1 to 1 that hosts should have initially, -2 for random
-set SYM_INT -2                 # Interaction value from -1 to 1 that symbionts should have initially, -2 for random
-set GRID_X 100                 # Width of the world, just multiplied by the height to get total size
-set GRID_Y 100                 # Height of world, just multiplied by width to get total size
-set POP_SIZE -1                # Starting size of the host population, -1 for full starting population
-set UPDATES 1001               # Number of updates to run before quitting
-set RES_DISTRIBUTE 100         # Number of resources to give to each host each update if they are available
-set LIMITED_RES_TOTAL -1       # Number of total resources available over the entire run, -1 for unlimited
-set HORIZ_TRANS 1              # Should non-lytic horizontal transmission occur? 0 for no, 1 for yes
-set HOST_REPRO_RES 1000        # How many resources required for host reproduction
-set SYM_HORIZ_TRANS_RES 100    # How many resources required for symbiont non-lytic horizontal transmission
-set GRID 0                     # Do offspring get placed immediately next to parents on grid, same for symbiont spreading
-set EFFICIENT_SYM 0            # Do you want symbionts that also have an efficiency value that evolves
-set EFFICIENCY_MUT_RATE -1     # The horizontal transmission mutation rate of the efficiency trait in symbionts, -1 if same as HORIZ_MUT_RATE
+set SEED 17                  # What value should the random seed be? If seed <= 0, then it is randomly re-chosen.
+set COMPETITION_MODE 0       # Should a competition between two types be conducted? (Which is specified in the code)
+set DATA_INT 100             # How frequently, in updates, should data print?
+set MUTATION_SIZE 0.002      # Standard deviation of the distribution to mutate by
+set HORIZ_MUTATION_SIZE -1   # Standard deviation of the distribution to mutate by for horizontal transmission, if -1 MUTATION_SIZE used
+set MUTATION_RATE 1          # Value 0 to 1 of probability of mutation
+set HORIZ_MUTATION_RATE -1   # Value 0 to 1 of probability of mutation for horizontal transmission, if -1 MUTATION_RATE used
+set SYNERGY 5                # Amount symbiont's returned resources should be multiplied by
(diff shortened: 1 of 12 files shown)


source/Sgghost.h
```diff
@@ -0,0 +1,139 @@
...
+
+  void Process(size_t location) {
+    //Currently just wrapping to use the existing function
+    double resources = my_world->PullResources();
+    DistribResources(resources);
+    // Check reproduction
+    if (GetPoints() >= my_config->HOST_REPRO_RES() && repro_syms.size() == 0) {  // if host has more points than required for repro
+        // will replicate & mutate a random offset from parent values
+        // while resetting resource points for host and symbiont to zero
+        emp::Ptr<sggHost> host_baby = emp::NewPtr<sggHost>(random, my_world, my_config, GetIntVal());
```
emilydolson, line 102: Is the type of `host_baby` the only reason we can't use the version of `Process` implemented in `Host`? That seems unfortunate, but all of the solutions are probably more trouble than they're worth.

source/sggsym.h
```diff
@@ -0,0 +1,116 @@
...
+
+  void mutate(){
+    // double pre_value = interaction_val;
+    if (random->GetDouble(0.0, 1.0) <= mut_rate) {
+      interaction_val += random->GetRandNormal(0.0, mut_size);
+      if(interaction_val < -1) interaction_val = -1;
+      else if (interaction_val > 1) interaction_val = 1;
+    }
+    if (random->GetDouble(0.0, 1.0) <= mut_rate) {
+      double mutate_size = random->GetRandNormal(0.0, mut_size);
```
emilydolson, line 35: You can go back to doing this all on one line (`sgg_donate += random->GetRandNormal(0.0, mut_size);`. We just had it this way for debugging

source/sggsym.h
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  void HorizMutate(){
+    // double pre_value = interaction_val;
+    if (random->GetDouble(0.0, 1.0) <= ht_mut_rate) {
+      interaction_val += random->GetRandNormal(0.0, ht_mut_size);
+      if(interaction_val < -1) interaction_val = -1;
+      else if (interaction_val > 1) interaction_val = 1;
+    }
+    if (random->GetDouble(0.0, 1.0) <= mut_rate) {
```
emilydolson, line 52: Should this be `ht_mut_rate` instead of `mut_rate`?

source/sggsym.h
```diff
@@ -0,0 +1,116 @@
...
+      sgg_donate += random->GetRandNormal(0.0, ht_mut_size);
+      if(sgg_donate < 0) sgg_donate = 0;
+      else if (sgg_donate > 1) sgg_donate = 1;
+    }
+  }
+
+  double ProcessPool(){
+    double symdonation = Getdonation();
+    double symPortion = GetPoints();
+    double hostreturn = symdonation*symPortion;
```
emilydolson, line 62: Why are these multiplied?

source/sggsym.h
```diff
@@ -0,0 +1,116 @@
...
+
+  emp::Ptr<Organism> reproduce() {
+    emp::Ptr<SggSymbiont> sym_baby = emp::NewPtr<SggSymbiont>(*this); //constructor that takes parent values
+    sym_baby->SetPoints(0);
+    sym_baby->mutate();
+    //mutate(); //mutate parent symbiont
+    return sym_baby;
+  }
+
+  void VerticalTransmission(emp::Ptr<Organism> host_baby) {
```
emilydolson, line 95: this looks like it's the same as in the base class, so you should be able to get rid of it here.

source/sggsym.h
```diff
@@ -0,0 +1,116 @@
...
+    if(my_world->WillTransmit()){
+      emp::Ptr<Organism> sym_baby = reproduce();
+      host_baby->AddSymbiont(sym_baby);
+    }
+  }
+
+};
+
+std::string PrintSym(emp::Ptr<SggSymbiont>  org){
```
emilydolson, line 104: This also looks the same as in the base class

SymSettings.cfg
```diff
@@ -1,6 +1,31 @@
 ### MAIN ###
```
emilydolson, line 1: The easiest way to fix the merge conflict in this file is probably to make the rest of the merges, compile symbulation, and regenerate this settings file from scratch

source/SymWorld.h
```diff
@@ -486,14 +586,20 @@ class SymWorld : public emp::World<Organism>{
 
     //TODO: put in fancy scheduler at some point
     emp::vector<size_t> schedule = emp::GetPermutation(GetRandom(), GetSize());
-
+    int n=0;
     // divvy up and distribute resources to host and symbiont in each cell
     for (size_t i : schedule) {
+<<<<<<< HEAD
       if (IsOccupied(i) == false){ continue;} // no organism at that cell
```
emilydolson, line 593: Looks like you can just stick with the original here

source/SymWorld.h
```diff
@@ -486,14 +586,20 @@ class SymWorld : public emp::World<Organism>{
 
     //TODO: put in fancy scheduler at some point
     emp::vector<size_t> schedule = emp::GetPermutation(GetRandom(), GetSize());
-
+    int n=0;
```
emilydolson, line 589: What is n doing?

source/SymWorld.h
```diff
@@ -34,7 +34,7 @@ class SymWorld : public emp::World<Organism>{
   emp::Ptr<emp::DataMonitor<double>> data_node_hosted_syms;
   emp::Ptr<emp::DataMonitor<double>> data_node_free_syms;
   emp::Ptr<emp::DataMonitor<int>> data_node_cfu;
-
+  emp::Ptr<emp::DataMonitor<double>> data_node_sgg;
```
emilydolson, line 37: Might be worth adding `emp::data::Histogram` to the template for this DataNode so that it tracks the full distribution of donation values (like the DataNodes for interactions between hosts and symbionts)

source/SymWorld.h
```diff
@@ -205,6 +206,45 @@ class SymWorld : public emp::World<Organism>{
 
     return file;
   }
+  emp::DataFile & SetupSggSymIntValFile(const std::string & filename) {
```
emilydolson, line 209: For this and the host equivalent, currently you've got it set up so that you're printing out a lot of duplicate information between files (since you're also setting up the standard data files). If you take my reccomendation above to track the full histogram for the donation value, I'd reccomend printing the histbins from that DataNode here, instead of from the SymInt DataNode (since you'll already have those in the sym file).

source/SymWorld.h
```diff
@@ -240,6 +280,44 @@ class SymWorld : public emp::World<Organism>{
     file.AddHistBin(node, 19, "Hist_0.9", "Count for histogram bin 0.9 to 1.0");
 
 
+    file.PrintHeaderKeys();
+
+    return file;
+  }
+  emp::DataFile & SetupSggHostIntValFile(const std::string & filename) {
```
emilydolson, line 287: I don't think you need this function, since there isn't any new host-specific data associated with the SGG stuff

emilydolson (changes requested):
This is looking pretty good overall! I made some small in-line comments about things to fix or that I have questions about.

A couple over-arching comments:
- It might be worth doing a find and replace to swap `sgg` with `pgg`, because this kind of dynamic is much more commonly referred to as a "public goods game" than as a "social goods game", although as far as I know they are technically synonyms
- Obviously, the merge conflict will need to be fixed. Looks like it shouldn't be too bad, though. Once the merged version is pushed, the test coverage will run and we can make sure your tests aren't missing anything.

PR #60, opened 2021-07-20
zhen sgg again 2

renztim:


emilydolson:
Looks like your new tests aren't getting run - make sure to add your new files to `source/catch/main.cc`

PR #62, opened 2021-07-21
zhen sgg 4

renztim:


source/Sgghost.h
```diff
@@ -0,0 +1,139 @@
...
+        SetResInProcess(sym_piece - hostDefense);
+      }
+      else if(hostIntVal >= 0){
+        hostDonation = hostIntVal * sym_piece;
+        SetResInProcess(sym_piece - hostDonation);
+      }
+      double sym_return = syms[i]->ProcessResources(hostDonation);
+      this->AddPoints(sym_return + GetResInProcess());
+      SetResInProcess(0);
+      double hostPool = syms[i]->ProcessPool();
```
emilydolson, line 76: For maintainability (since only the last couple lines of this for loop are different from the base class), it would probably be better to have this call the version of DistribResources in the base class and then make a new loop for the pool stuff. The more we can avoid duplicated code in a project like this the better. So the function would look like this:

```
void DistribResources(double resources) {
	Host::DistribResources(resources);
	for(size_t i=0; i < syms.size(); i++){
		double hostPool = syms[i]->ProcessPool();
		this->AddPool(hostPool);
	}
	if(syms.size()>0){this->DistribPool();}

} //end DistribResources
```

source/Sgghost.h
```diff
@@ -0,0 +1,139 @@
...
+    //to do: marginal return
+    int num_sym = syms.size();
+    double sym_piece = (double) sourcepool / num_sym;
+    for(size_t i=0; i < syms.size(); i++){
+        syms[i]->AddPoints(sym_piece);
+    }
+    this->SetPool(0);
+  }
+
+  void Process(size_t location) {
```
emilydolson, line 94: Looks like this function is getting run by any of the tests. Try to add at least one that runs it.

source/sggsym.h
```diff
@@ -0,0 +1,108 @@
...
+  SggSymbiont(const SggSymbiont &) = default;
+  SggSymbiont(SggSymbiont &&) = default;
+  SggSymbiont() = default;
+
+
+  SggSymbiont & operator=(const SggSymbiont &) = default;
+  SggSymbiont & operator=(SggSymbiont &&) = default;
+
+
+  double Getdonation() {return sgg_donate;}
```
emilydolson, line 24: Could you rename these functions to GetDonation and SetDonation for consistency with the style of the rest of the code?

source/sggsym.h
```diff
@@ -0,0 +1,108 @@
...
+  void Setdonation(double _in) {sgg_donate = _in;}
+
+  void mutate(){
+    // double pre_value = interaction_val;
+    if (random->GetDouble(0.0, 1.0) <= mut_rate) {
```
emilydolson, line 29: Could replace this if statement with a call to `Symbiont::mutate()` to reduce the amount of duplicate code (and insure that future changes to the base mutate function get reflected here)

source/sggsym.h
```diff
@@ -0,0 +1,108 @@
...
+      else if (sgg_donate > 1) sgg_donate = 1;
+    }
+    //if((pre_value*interaction_val) < 0) {
+    //  std::cout << "switched2!" << std::endl;
+    //}
+  }
+
+  void HorizMutate(){
+    // double pre_value = interaction_val;
+    if (random->GetDouble(0.0, 1.0) <= ht_mut_rate) {
```
emilydolson, line 46: This too

source/sggsym.h
```diff
@@ -0,0 +1,108 @@
...
+    sym_baby->mutate();
+    //mutate(); //mutate parent symbiont
+    return sym_baby;
+  }
+
+
+std::string PrintSym(emp::Ptr<SggSymbiont>  org){
+  if (org->GetPoints() < 0) return "-";
+  double out_val = org->GetIntVal();
+
```
emilydolson, line 98: Might want to output the donation value too

emilydolson (changes requested):
Looking good! Almost there! I made a few more in-line comments, but they're mostly pretty minor polishing things.

I still think it might be worth switching the SGG (for "Social Goods Game") prefix to PGG (for "Public Goods Game") because that is the more commonly used term. It should be a not-too-bad find and replace. What do you think, @anyaevostinar?

Is the other pull request (#59) something different than this one or can it get closed? If possible it would be great to keep all of this in a single pull request, for organization purposes (and so I can more easily see which code I've already looked at). Any commits that you put on the main branch of your fork should end up here and all the checks on this pull request will automatically update. Let me know if you have any questions!

One other question for @anyaevostinar about the contribution process for Symbulation: once I've approved this pull request, do you want to also review it before it gets merged?

source/ConfigSetup.h
```diff
@@ -43,6 +43,10 @@ EMP_BUILD_CONFIG(SymConfigBase,
     VALUE(FREE_LIVING_SYMS, bool, 0, "Are symbionts able to live outside of the host?"),
     VALUE(MOVE_FREE_SYMS, bool, 0, "Should the free living phage move around in the world? 0 for no, 1 for yes"),
     
+    GROUP(PGG,"Public Goods Game Settings"),
```
anyaevostinar, line 46: This should go after the file_path and file_name since otherwise they will be lumped into the PGG group

source/Organism.h
```diff
@@ -149,6 +149,27 @@ class Organism {
     std::cout << "GetLysogeny called from Organism" << std::endl;
     throw "Organism method called!";}
 
+  //Public goods game host functions 
+  virtual void SetPool() {
+    std::cout << "Set resources pool from Organism" << std::endl;
```
anyaevostinar, line 154: These should each be the exact name of the method to make it easier to track down if/when it happens

source/PggHostPggSymUnitTest.test.cc
```diff
@@ -0,0 +1,449 @@
...
+    emp::Ptr<Organism> h = new PggHost(random, &w, &config);
+    PGGSymbiont * s = new PGGSymbiont(random, &w, &config, int_val);
+
+    s->SetHost(h);
+
+    REQUIRE(s->GetHost() == h);
+
+}
+
+TEST_CASE("PGGHost SetSymbionts") {
```
anyaevostinar, line 20: It doesn't look like you have your own new methods for these, so I'm not sure why you need tests for them if you are just using the Host and Symbiont methods

source/WorldSetup.cc
```diff
@@ -34,22 +36,46 @@ void worldSetup(emp::Ptr<SymWorld> world, emp::Ptr<SymConfigBase> my_config) {
...
+        world->AddOrgAt(new_org, world->size());
+      }
+      //world.Inject(*new_org);
     }
-    //Currently hacked because there isn't an AddOrg function, but there probably should be
-    if(my_config->GRID()) {
-      world->AddOrgAt(new_org, emp::WorldPosition(world->GetRandomCellID()));
-    } else {
-      world->AddOrgAt(new_org, world->size());
+  } else{
```
anyaevostinar, line 60: This is a whole lot of repeated code, you could probably just have the emp::Ptr<Host> new org in an if statement instead from the looks of it.

anyaevostinar (changes requested):
I've put in some comments focused on reducing repeated code/bloat

SymSettings.cfg
```diff
@@ -32,11 +32,20 @@ set LYSIS_CHANCE -1        # Chance of lysis vs. lysogeny for starting populatio
...
+set PHAGE_EXCLUDE 0        # Do phage have a decreased chance of getting into the host if there is already a lytic phage?
+set FREE_LIVING_SYMS 0     # Are symbionts able to live outside of the host?
+set MOVE_FREE_SYMS 0       # Should the free living phage move around in the world? 0 for no, 1 for yes
+
+### SGG ###
+# Social Goods Game Settings
+
+set SGG_DONATE 0.01                            # Ratio of symbionts‘ energy to sgg pool that experiment should start with
+set SGG 1                                      # whether have social goods game among syms
+set FILE_PATH source/sgg_end_to_end_test_data  # Output file path
```
anyaevostinar, line 49: These are still changed from the default: "FILE_PATH" and "FILE_NAME"

SymSettings.cfg
```diff
@@ -32,11 +32,21 @@ set LYSIS_CHANCE -1        # Chance of lysis vs. lysogeny for starting populatio
...
 set PROPHAGE_LOSS_RATE 0   # Rate at which infected lysogens become re-susceptible to new phage
-set SYM_LYSIS_RES 1        # How many resources required for symbiont to create offspring for lysis each update
+set SYM_LYSIS_RES 1      # How many resources required for symbiont to create offspring for lysis each update
 set START_MOI 1            # Ratio of symbionts to hosts that experiment should start with
-set FILE_PATH              # Output file path
-set FILE_NAME _data_       # Root output file name
+set PHAGE_EXCLUDE 0        # Do phage have a decreased chance of getting into the host if there is already a lytic phage?
+set FREE_LIVING_SYMS 0     # Are symbionts able to live outside of the host?
+set MOVE_FREE_SYMS 0       # Should the free living phage move around in the world? 0 for no, 1 for yes
+set FILE_PATH source/sgg_end_to_end_test_data  # Output file path
```
anyaevostinar, line 43: These two still need to be changed back

source/Pggendtoend.test.cc
```diff
@@ -0,0 +1,131 @@
...
+}//void Test(.........)
+
+/*Test(std::string expected_result_file, int seed = 10, int data_int = 100, double mutation_rate = 0.002, 
+    double synergy = 5, double vertical_transmission = 1, double host_int = 0, 
+    double sym_int = 0,  int grid_x = 5, int grid_y = 5, int updates = 1, 
+    int sym_limit = 1, bool lysis = 0, bool horiz_trans = 0, int burst_size = 10, 
+    int burst_time = 10, double host_repro_res = 1000,double sym_lysis_res = 1, 
+	  double sym_horiz_trans_res = 100, int start_moi = 1, bool grid = 0)
+*/
+TEST_CASE( "End To End" ) {
```
anyaevostinar, line 126: It doesn't look like data files are actually being added to check against, so I don't think this test will do anything. I also think we are going to try to move away from this kind of test since it is difficult to maintain, so I would prefer you remove it.

source/Pgghost.h
```diff
@@ -0,0 +1,139 @@
...
+
+
+public:
+  PggHost(emp::Ptr<emp::Random> _random, emp::Ptr<SymWorld> _world, emp::Ptr<SymConfigBase> _config,
+  double _intval =0.0, emp::vector<emp::Ptr<Organism>> _syms = {},
+  emp::vector<emp::Ptr<Organism>> _repro_syms = {},
+  std::set<int> _set = std::set<int>(),
+  double _points = 0.0) : Host(_random, _world, _config, _intval,_syms, _repro_syms, _set, _points) { 
+    if ( _intval > 1 || _intval < -1) {
+       throw "Invalid interaction value. Must be between -1 and 1";  // Exception for invalid interaction value
```
anyaevostinar, line 25: You don't need this line in here since the Host constructor will do it, it's just repeated code.

source/Pgghost.test.cc
```diff
@@ -0,0 +1,112 @@
...
+    double default_pool = 0.0;
+    REQUIRE(h1->GetPool() == default_pool);
+
+    PggHost * h2 = new PggHost(random, &w, &config);
+    h2->SetPool(pool);
+    double expected_pool = 1;
+    REQUIRE(h2->GetPool() == expected_pool);
+
+}
+TEST_CASE("Pgghost SetIntVal, GetIntVal") {
```
anyaevostinar, line 21: Why do these need to be tested? Aren't they just using the Host's methods?

anyaevostinar (changes requested):
Still some repeated code and tests that should get resolved

source/WorldSetup.cc
emilydolson: This looks like it will be easier to fix once #67 is resolved - should we hold off on fixing it until #67 is fixed?

anyaevostinar: I think it should be easy in this case since it looks like the only line that is different between these blocks is " emp::Ptr<PggHost> new_org;" versus "emp::Ptr<Host> new_org;" so the MakeNew wouldn't be needed anyway

emilydolson: Sorry if I'm being dumb here, but if he puts it in an if statement won't it go out of scope once the if statement ends?

source/Pgghost.h
```diff
@@ -0,0 +1,135 @@
...
+  PggHost(const PggHost &) = default;
+  PggHost(PggHost &&) = default;
+  PggHost() = default;
+ 
+  double GetPool() {return sourcepool;}
+  void SetPool(double _in) {sourcepool= _in;}
+  void AddPool(double _in) {sourcepool += _in;}
+
+
+  void DistribResources(double resources) {
```
emilydolson, line 36: I think I had a comment about this before, but it must have gotten lost. You can replace most of this function with a call to Host::DistribResources(resources); Then you just need to add a for loop after it to handle the pool stuff (which is the only difference between this function and the Host's version)

emilydolson:
I cleaned up a couple minor differences that were making the git history messy and did one last readthrough of the changes. The only definite remaining issue that I see is the duplicated code in DistribResources, and possibly the thing in WorldSetup.cc

source/WorldSetup.cc
renztim: I tried to put the emp::Ptr new_org in an if statment before but it returns an error said that new_org was not decleared in the scope.

emilydolson:
Great work! This addresses all of my comments, so I'm going to go ahead and approve. I think the only possible outstanding issue is that if statement.

emilydolson (approved):

anyaevostinar (approved):
Great job!

PR #77, opened 2021-08-06
Lysogeny dev

acameron13:
Implemented spontaneous induction, added a data file to track it, and condensed the lysis related data files


source/ConfigSetup.h
```diff
@@ -32,9 +32,11 @@ EMP_BUILD_CONFIG(SymConfigBase,
 
     GROUP(LYSIS, "Lysis Settings, coming soon to the GUI!"),
     VALUE(LYSIS_CHANCE, double, -1, "Chance of lysis vs. lysogeny for starting population of phage, -1 for random distribution"),
+    VALUE(CHANCE_OF_INDUCTION, double, 0, "Chance of induction for starting lysogenic phage, -1 for random distribution"),
     VALUE(MUTATE_LYSIS_CHANCE, int, 0, "Should the chance of lysis mutate? 0 for no, 1 for yes"),
     VALUE(SYM_LIMIT, int, 1, "Number of symbiont allowed to infect a single host"),
     VALUE(LYSIS, bool, 0, "Should lysis occur? 0 for no, 1 for yes"),
+    VALUE(MUTATE_INDUCTION_CHANCE, int, 0, "Should the chance of Induction mutate? 0 for no, 1 for yes"),
```
anyaevostinar, line 39: Should be a bool, not int

source/HostPhageUnitTest.test.cc
```diff
@@ -230,3 +230,94 @@ TEST_CASE("Host phage death and removal from syms list"){
...
+    config.LYSIS(1); //phage process only happens when lysis is enabled
+    config.GRID_X(2); 
+    config.GRID_Y(1);
+    config.SYM_LIMIT(2);
+    int location = 0;
+
+    WHEN("The phage chooses to Horizontally Transmit"){
+        config.LYSIS_CHANCE(1.0); //100% chance of lysis, 0% chance of lysogeny
+        config.SYM_LYSIS_RES(5);
+        config.BURST_TIME(10);
```
anyaevostinar, line 249: This 10 is used in two places, so it'd be better for it to be a variable to make it clear that they are linked

source/HostPhageUnitTest.test.cc
```diff
@@ -230,3 +230,94 @@ TEST_CASE("Host phage death and removal from syms list"){
...
+
+    config.LYSIS(1); //phage process only happens when lysis is enabled
+    config.GRID_X(2); 
+    config.GRID_Y(1);
+    config.SYM_LIMIT(2);
+    int location = 0;
+
+    WHEN("The phage chooses to Horizontally Transmit"){
+        config.LYSIS_CHANCE(1.0); //100% chance of lysis, 0% chance of lysogeny
+        config.SYM_LYSIS_RES(5);
```
anyaevostinar, line 248: This 5 is used twice and so should be a variable to make it clear that later in the test, the 5's are linked.

source/HostPhageUnitTest.test.cc
```diff
@@ -230,3 +230,94 @@ TEST_CASE("Host phage death and removal from syms list"){
...
+                    double expected_burst_timer = 1.5306015114;
+                    REQUIRE(p->GetBurstTimer() == Approx(expected_burst_timer));
+                    REQUIRE(size(h->GetReproSymbionts()) == repro_syms_size_pre_process);
+                    REQUIRE(p->GetPoints() == expected_points);
+                }
+            }
+
+            WHEN("The phage has enough resources to reproduce"){
+                double expected_repro_syms_size_post_process = size(h->GetReproSymbionts()) + 1; //one offspring created
+                double orig_points = 5.0;
```
anyaevostinar, line 306: This is the other 5 that should be a variable to make it clear you are giving them just enough resources to reproduce

source/Phage.h
```diff
@@ -65,11 +80,16 @@ class Phage: public Symbiont {
...
     sym_lysis_res = my_config->SYM_LYSIS_RES();
     lysis_enabled = my_config->LYSIS();
+    mutate_chance_of_induction = my_config->MUTATE_INDUCTION_CHANCE();
     mutate_chance_of_lysis = my_config->MUTATE_LYSIS_CHANCE();
     chance_of_lysis = my_config->LYSIS_CHANCE();
+    induction_chance = my_config->CHANCE_OF_INDUCTION();
     if(chance_of_lysis == -1){
       chance_of_lysis = random->GetDouble(0.0, 1.0);
     }
+    if(induction_chance == -1){
```
anyaevostinar, line 90: This is easy to add a test for, so you should

source/Phage.h
```diff
@@ -233,6 +273,44 @@ class Phage: public Symbiont {
...
   }
+
+  /**
+   * Input: location of the phage attempting to horizontally transmit
+   * 
+   * Output: None
+   * 
+   * Purpose: To allow lytic/induced phage to lyse their host 
+   */
+  void HorizontalTransmission(size_t location) { //Horizontal transmission process after lysis or induction are called
```
anyaevostinar, line 284: I'm not convinced that HorizontalTransmission is the best name for this method. Isn't it just about Lysis (since there are other ways that horizontal transmission can happen with phage theoretically)? Or maybe it should actually be two different methods (LysisBurst and LysisStep maybe?) and the burst timer if statement could stay in process. 

The main concern I have is that the code in Process is confusing since it reads "not lysogeny, so horizontal transmission" which hides too many of the details of what is actually going on. It should read "not lysogeny, so check the burst timer and either lyse or make another offspring for eventual lysis"

source/Phage.h
```diff
@@ -281,43 +358,20 @@ class Phage: public Symbiont {
...
-        }
+        HorizontalTransmission(location);
       }
       else if(lysogeny){ //phage has chosen lysogeny
-        //check if the phage's host should become susceptible again
-        if(random->GetDouble(0.0, 1.0) <= my_config->PROPHAGE_LOSS_RATE()){
+        double rand_chance = random->GetDouble(0.0, 1.0);
+        if (rand_chance <= induction_chance){//phage has chosen to induce and perform lysis
+          lysogeny = false;
+          HorizontalTransmission(location);
```
anyaevostinar, line 367: Does it work to not call HorizontalTransmission immediately and instead just set lysogeny to false and wait until the next update to actually start the lysis process?

source/Phage.test.cc
```diff
@@ -283,92 +313,60 @@ TEST_CASE("Phage process"){
...
+            THEN("The phage doesn't induce and is set to dead"){
                 REQUIRE(p->GetDead() == expected_dead);
             }
 
             h.Delete();
         }
-        
-    }
 
-    WHEN("The phage chooses lysis"){
```
anyaevostinar, line 295: Was this test moved, if so where? It looks like this new test is testing something different, so I want to make sure we didn't lose the "When the phage chooses lysis" test

anyaevostinar: Looks like it was moved to HostPhageUnitTest and the name changed

anyaevostinar (changes requested):
A few changes requested

source/HostPhageUnitTest.test.cc
```diff
@@ -230,3 +230,88 @@ TEST_CASE("Host phage death and removal from syms list"){
...
+    config.LYSIS(1); 
+    config.GRID_X(2); 
+    config.GRID_Y(1);
+    config.SYM_LIMIT(2);
+    int location = 0;
+
+    double int_val = 0;
+    emp::Ptr<Phage> p = new Phage(random, world, &config, int_val);
+    
+    //create two hosts and add both to world as neighbors
```
anyaevostinar, line 249: Some "WHEN" and "THEN" additions will make this test easier to understand

anyaevostinar (approved):
Looks great!

PR #103, opened 2021-11-14
Death by old age and independent host mutation

K-Johnson-Horrigan:


source/Symbiont.test.cc
```diff
@@ -538,3 +538,37 @@ TEST_CASE("Symbiont ProcessResources"){
...
+    config.SYM_AGE_MAX(2);
+
+    WHEN ("A free-living symbiont reaches its maximum age"){
+      config.FREE_LIVING_SYMS(1);
+      Symbiont * s = new Symbiont(random, &w, &config, 1);
+      w.AddOrgAt(s, 1);
+      THEN("The symbiont dies and gets removed from the world"){
+        REQUIRE(w.GetNumOrgs() == 1);
+        REQUIRE(s->GetDead() == false);
+        w.Update(); //sym goes from age 1->2
```
anyaevostinar, line 556: If we had a getAge method in the symbiont, that would be helpful to check here that the age goes up correctly

source/Symbiont.test.cc
```diff
@@ -538,3 +538,37 @@ TEST_CASE("Symbiont ProcessResources"){
...
+        w.Update(); //sym goes from age 2->3, gets set to dead
```
anyaevostinar, line 557: Should probably check that it's getDead is true?

source/Symbiont.test.cc
```diff
@@ -538,3 +538,37 @@ TEST_CASE("Symbiont ProcessResources"){
...
+    }
+    WHEN ("A hosted symbiont reaches its maximum age"){
+      Symbiont * s = new Symbiont(random, &w, &config, 1);
+      Host * h = new Host(random, &w, &config, 1);
+      h->AddSymbiont(s);
+      THEN("It dies and gets removed from its host")
+        REQUIRE(h->HasSym() == true);
+        h->Process(1);
+        h->Process(1); //sym reaches age 2 and is set to dead
+        //REQUIRE(h->HasSym() == true);
```
anyaevostinar, line 570: Why is this commented out?

anyaevostinar (changes requested):
Looking good, just a few things that need changing. Main question is if we can remove dead organisms right away instead of at the beginning of the next cycle

anyaevostinar (approved):
I've made the required changes

PR #104, opened 2021-11-16
Subclass-specific functions and variables

K-Johnson-Horrigan:


source/efficient_mode/EfficientSymbiont.h
```diff
@@ -120,43 +120,14 @@ class EfficientSymbiont: public Symbiont {
 
 
   /**
-   * Input: None
+   * Input: None.
    *
-   * Output: An Empirical pointer to a newly created organism
+   * Output: A new bacterium with same properties as this bacterium.
```
anyaevostinar, line 125: This should probably say EfficiencySymbiont?

anyaevostinar (changes requested):
A few changes to be made still

PR #105, opened 2021-12-02
mutate() param and GetDoEcto earlier return

K-Johnson-Horrigan:
Added Organism mutate() parameter to subclass mutate() functions and modified GetDoEctosymbiosis() so that it returns false if the config option for ectosymbiosis is off, prior to determining if other factors are present.

### DIFF
--- a/source/default_mode/Host.h
+++ b/source/default_mode/Host.h
@@ -1,6 +1,6 @@
 #ifndef HOST_H
 #define HOST_H
- 
+
 #include "../../../Empirical/include/emp/math/Random.hpp"
 #include "../../../Empirical/include/emp/tools/string_utils.hpp"
 #include <set>
@@ -390,7 +390,7 @@ public:
    *
    * Purpose: Increments age by one and kills it if too old.
    */
-  void growOlder(){
+  void GrowOlder(){
     age = age + 1;
     if(age > my_config->HOST_AGE_MAX() && my_config->HOST_AGE_MAX() > 0){
       SetDead();
@@ -533,7 +533,7 @@ public:
    * Purpose: To mutate a host's interaction value. This is called on newly generated
    * hosts to allow for evolution to occur.
    */
-  void mutate(){
+  void mutate(std::string mode = "vertical"){
     double mutation_size = my_config->HOST_MUTATION_SIZE();
     if (mutation_size == -1) mutation_size = my_config->MUTATION_SIZE();
     double mutation_rate = my_config->HOST_MUTATION_RATE();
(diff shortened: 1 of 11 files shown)


source/default_mode/Host.h
```diff
@@ -533,7 +533,7 @@ class Host: public Organism {
    * Purpose: To mutate a host's interaction value. This is called on newly generated
    * hosts to allow for evolution to occur.
    */
-  void mutate(){
+  void mutate(std::string mode = "vertical"){
```
anyaevostinar, line 536: This should default to "none" for hosts since they don't have vertical transmission

source/efficient_mode/EfficientSymbiont.h
```diff
@@ -91,7 +91,7 @@ class EfficientSymbiont: public Symbiont {
    * Purpose: Mutating the efficiency of an efficient symbiont based upon the config
    * setting for mutation size.
    */
-  void mutate(){
+  void mutate(std::string mode = "vertical"){
     Symbiont::mutate();
```
anyaevostinar, line 95: We need to set up the mutate function to work for horizontal transmission mutation as well in these subclasses

source/lysis_mode/Bacterium.h
```diff
@@ -121,7 +121,7 @@ class Bacterium : public Host {
    * chosen from a normal distribution centered at 0, with a standard deviation that
    * is equal to the mutation size. Bacterium mutation can be turned on or off.
    */
-  void mutate() {
+  void mutate(std::string mode = "vertical") {
     Host::mutate();
```
anyaevostinar, line 125: Change to 'none' for host subclasses

source/lysis_mode/Phage.h
```diff
@@ -284,7 +284,7 @@ class Phage: public Symbiont {
    * deviation that is equal to the mutation size. Phage mutation can be
    * on or off.
    */
-  void mutate() {
+  void mutate(std::string mode = "vertical") {
    Symbiont::mutate();
```
anyaevostinar, line 288: Change so that it processes 'horizontal' argument correctly

source/pgg_mode/Pggsym.h
```diff
@@ -102,7 +102,7 @@ class PGGSymbiont: public Symbiont {
    *
    * Purpose:
    */
-  void mutate(){
+  void mutate(std::string mode = "vertical"){
```
anyaevostinar, line 105: Change to work with 'horizontal' correctly

anyaevostinar (commented):
Thank you for tracking down those issues! We just need to fix it so that the horizontal transmission mutation rate is actually used properly when the argument is passed as horizontal, which I can do today

PR #106, opened 2022-01-07
Catch tag fixes and start to more quickstart guides

acameron13:


source/test/default_mode_test/Host.test.cc
```diff
@@ -220,7 +220,7 @@ TEST_CASE("DistributeResources", "[default][efficient][lysis][pgg]") {
     }
 }
 
-TEST_CASE("Phage Exclude", "[default][efficient][lysis][pgg]") {
+TEST_CASE("Phage Exclude", "[default]") {
```
anyaevostinar, line 223: Phage exclude should probably be moved to Bacterium

anyaevostinar (approved):

PR #115, opened 2022-01-25
makeNew & reproduce

acameron13:
Changed the makeNew and reproduce functions in the following ways:
makeNew does not use any copy constructors, and instead creates a new organism and subsequently adds all relevant genome values to the new organism. makeNew does not do any mutating.
reproduce calls the makeNew function, mutates the offspring genome, and then sets the parent's points to the correct amount. 

This refactor reduces the amount of repeated code since each organism now only needs a makeNew function, not a reproduce function with repeated functionality.

Also included in this pull request is moving source/SymWorld.h into source/default_mode/SymWorld.h and removing source/WorldSetup.cc altogether.


anyaevostinar (approved):
Really lovely additions by the way, the inclusion of the comments and organization was wonderful to see

PR #125, opened 2022-04-05
PGG names and documentation

K-Johnson-Horrigan:
Modified PGG class names to match the format <PGGName> and added documentation to PGG variables/methods.


source/pgg_mode/PGGWorld.h
```diff
@@ -20,44 +20,49 @@ class PggWorld : public SymWorld {
...
+    * symbionts at various donation values.
+    */
   emp::DataFile & SetupPGGSymIntValFile(const std::string & filename) {
     auto & file = SetupFile(filename);
     auto & node1 = GetSymCountDataNode();
     auto & node2 = GetCountFreeSymsDataNode();
     auto & node3 = GetCountHostedSymsDataNode();
     auto & node4 = GetPGGDataNode();
-    node4.SetupBins(0, 1.1, 11); //Necessary because range exclusive
+    node4.SetupBins(0.0, 1.1, 10); //Necessary because range exclusive
```
anyaevostinar, line 48: I think this should still have 11 as the final argument so that the bins are each .1 in size

anyaevostinar (changes requested):
Looking pretty good, just concerned about the change to the histogram bin in a comment

anyaevostinar (approved):
Awesome!

PR #126, opened 2022-04-06
Added submodules

acameron13:
Changed all file paths that include things from Empirical


anyaevostinar:
I think the github actions will need to be updated to make sure that they pull down Empirical correctly

emilydolson:
Yep, that's exactly right! I just sent @acameron13 a pull request that should fix it. Merging that pull request on her repo should merge the changes into this pull request (because I made the pull request to the branch that that this one is based off of).

source/test/lysis_mode_test/LysisDataNodes.test.cc
```diff
@@ -1,62 +1,62 @@
-#include "../../lysis_mode/Phage.h"
```
anyaevostinar, line 1: Did anything change in this file or is github just being weird?

anyaevostinar (commented):
Looks good, just want to check on if there is actually any change to the LysisDataNodes.test.cc file?

source/test/lysis_mode_test/LysisDataNodes.test.cc
acameron13: I am pretty sure it is just github being weird. I just merged by pulling down from your main and accepted all incoming changes, since my old test for that was just me messing around with histogram bins. But the old test isn't there at all any more

anyaevostinar (approved):

PR #130, opened 2022-04-11
Removing namespace

acameron13:
In all source/native files, I removed the line saying "using namespace std" and added "std::" in front of all functions in the standard library


anyaevostinar (approved):
Great, thank you for fixing this!

PR #141, opened 2022-05-25
Phage ProcessResources inheritance

K-Johnson-Horrigan:
Changing Symbiont.h ProcessResources to take a host parameter that defaults to my_host & changing Phage.h ProcessResources to take the same parameters.

### DIFF
--- a/source/default_mode/Symbiont.h
+++ b/source/default_mode/Symbiont.h
@@ -388,26 +388,17 @@ public:
     }
   }
 
-
   /**
-   * Input: The double representing the resources to be distributed to the symbionts
-   *
-   * Output: The double representing the host's resources
-   *
-   * Purpose: To process and distribute resources.
-  */
-  double ProcessResources(double hostDonation){
-    return ProcessResources(hostDonation, my_host);
-  }
-
-  /**
-   * Input: The double representing the resources to be distributed to the symbionts and the host from whom it comes
+   * Input: The double representing the resources to be distributed to the symbionts and the host from whom it comes; if no host is provided it will automatically use my_host
    *
    * Output: The double representing the host's resources
    *
    * Purpose: To process and distribute resources.
    */
-  double ProcessResources(double hostDonation, emp::Ptr<Organism> host){
+  double ProcessResources(double host_donation, emp::Ptr<Organism> host = nullptr){
+    if(host == nullptr){
+      host = my_host;
+    }
     double sym_int_val = GetIntVal();
     double sym_portion = 0;
     double host_portion = 0;
@@ -416,11 +407,11 @@ public:
     if (sym_int_val<0){
       double stolen = host->StealResources(sym_int_val);
       host_portion = 0;
-      sym_portion = stolen + hostDonation;
+      sym_portion = stolen + host_donation;
     }
     else if (sym_int_val >= 0){
-      host_portion = hostDonation * sym_int_val;
-      sym_portion = hostDonation - host_portion;
+      host_portion = host_donation * sym_int_val;
+      sym_portion = host_donation - host_portion;
     }
     AddPoints(sym_portion);
     return host_portion * synergy;
--- a/source/default_mode/Symbiont.h
+++ b/source/default_mode/Symbiont.h
@@ -389,7 +389,9 @@ public:
   }
 
   /**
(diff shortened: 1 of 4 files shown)


source/default_mode/Symbiont.h
```diff
@@ -388,26 +388,17 @@ class Symbiont: public Organism {
...
-   *
-   * Purpose: To process and distribute resources.
-  */
-  double ProcessResources(double hostDonation){
-    return ProcessResources(hostDonation, my_host);
-  }
-
   /**
-   * Input: The double representing the resources to be distributed to the symbionts and the host from whom it comes
+   * Input: The double representing the resources to be distributed to the symbionts and the host from whom it comes; if no host is provided it will automatically use my_host
```
anyaevostinar, line 392: Change to:

Input: The double representing the resources to be distributed to the symbionts and (optionally) the host from whom it comes; if no host is provided, the symbiont's host variable is used.

source/lysis_mode/Phage.h
```diff
@@ -346,16 +346,19 @@ class Phage: public Symbiont {
    *
    * Purpose: To allow a phage to steal or use donated resources from their host.
```
anyaevostinar, line 347: Should update the docstring to mention the optional host argument

source/test/lysis_mode_test/Phage.test.cc
```diff
@@ -662,6 +662,41 @@ TEST_CASE("Phage ProcessResources", "[lysis]"){
             bacterium.Delete();
         }
     }
+
+    THEN("Phage, not Symbiont, ProcessResources is used"){
```
anyaevostinar, line 666: This should be in BacteriumPhageUnitTest since it is using a Bacterium

source/test/lysis_mode_test/Phage.test.cc
```diff
@@ -662,6 +662,41 @@ TEST_CASE("Phage ProcessResources", "[lysis]"){
...
+      config.LYSIS(1);
+      config.LYSIS_CHANCE(0);
+      config.BENEFIT_TO_HOST(0);
+      config.SYNERGY(5);
+
+      emp::Ptr<Organism> phage = emp::NewPtr<Phage>(random, world, &config, int_val);
+      emp::Ptr<Host> bacterium = emp::NewPtr<Bacterium>(random, world, &config, int_val);
+      phage->UponInjection();
+
+      emp::Ptr<Organism> symbiont = emp::NewPtr<Symbiont>(random, world, &config, int_val);
```
anyaevostinar, line 678: I assume you had these to check what the symbiont's processresources would do, but they shouldn't be in this test. You could put in a comment about what the symbiont's process resources would result in and versus the phage's process resources, but the test should be just making the phage and bacterium and calling DistribResToSym and making sure that the amount the results is correct.

source/test/lysis_mode_test/Phage.test.cc
```diff
@@ -662,6 +662,41 @@ TEST_CASE("Phage ProcessResources", "[lysis]"){
...
+      emp::Ptr<Host> host = emp::NewPtr<Host>(random, world, &config, int_val);
+      REQUIRE(host->GetPoints() == bacterium->GetPoints());
+
+      double sym_piece = 20;
+      host->DistribResToSym(symbiont, sym_piece);
+      bacterium->DistribResToSym(phage, sym_piece);
+
+      double expected_host_points = 20;
+      double expected_bacterium_points = 16;
```
anyaevostinar, line 687: Let's definitely put in a comment explaining where the number 16 comes from

source/test/lysis_mode_test/Phage.test.cc
```diff
@@ -662,6 +662,41 @@ TEST_CASE("Phage ProcessResources", "[lysis]"){
...
+      double host_points = host->GetPoints();
+      double bacterium_points = bacterium->GetPoints();
+
+      REQUIRE(host_points != bacterium_points);
+      REQUIRE(host_points == expected_host_points);
+      REQUIRE(bacterium_points == expected_bacterium_points);
```
anyaevostinar, line 693: Should the phage's points be checked as well?

anyaevostinar (changes requested):
Fix looks good, just want to streamline the new test a bit

anyaevostinar (approved):
Awesome!

PR #145, opened 2022-06-14
No Mutation Updates

anyaevostinar:
Introducing functionality for a set number of updates to occur after the main experiment where no mutations happen, allowing for the dominant lineage(s) to become more apparent


K-Johnson-Horrigan (approved):
Looks good!

PR #149, opened 2022-06-17
Transmission data nodes

K-Johnson-Horrigan:
Three new data nodes + a file describing symbiont transmission:
- attempted horizontal transmission
- successful horizontal transmission (is not deleted when attempting to move into host/ free world cell)
- attempted vertical transmission

Some functions that handle symbiont movement now return positional data:
-  SymDoBirth() and MoveIntoNewFreeWorldPos() return emp::WorldPosition objects describing the symbiont's new location. 
- Host AddSymbiont() returns the location of the symbiont in the host vector (offset by 1).

Checker for WorldPosition object validity created; checks that the index/ID are within bounds if they are being used to describe free-living organism location.


source/default_mode/Host.h
```diff
@@ -449,17 +449,20 @@ class Host: public Organism {
   /**
    * Input: The pointer to the organism that is to be added to the host's symbionts.
    *
-   * Output: None
+   * Output: The int describing the symbiont's position ID, or 0 if it did not successfully
+   * get added to the host's list of symbionts.
```
anyaevostinar, line 453: Can syms not be at position 0? If so, should be specified since that's different than normal

source/default_mode/Symbiont.h
```diff
@@ -543,6 +551,10 @@ class Symbiont: public Organism {
       emp::Ptr<Organism> sym_baby = Reproduce();
       points = points - my_config->SYM_VERT_TRANS_RES();
       host_baby->AddSymbiont(sym_baby);
+
+      //vertical transmission data node
+      emp::DataMonitor<int>& data_node_attempts_verttrans = my_world->GetVerticalTransmissionAttemptCount();
```
anyaevostinar, line 556: Do we need to get it each time or can we make sure its initialized somewhere else and just use the instance variable?

source/default_mode/Symbiont.h
```diff
@@ -561,7 +573,16 @@ class Symbiont: public Organism {
         //points = points - my_config->SYM_HORIZ_TRANS_RES();
         SetPoints(0);
         emp::Ptr<Organism> sym_baby = Reproduce();
-        my_world->SymDoBirth(sym_baby, location);
+        emp::WorldPosition new_pos = my_world->SymDoBirth(sym_baby, location);
+
+        //horizontal transmission data nodes
+        emp::DataMonitor<int>& data_node_attempts_horiztrans = my_world->GetHorizontalTransmissionAttemptCount();
```
anyaevostinar, line 579: Ditto here, do we need to call the Get methods each time or can we use the instance variables when we're sure it has been created?

anyaevostinar (approved):

PR #156, opened 2022-07-07
Internal environment features, new squaring task

andersonm2:


source/sgp_mode/SquareTest.h
```diff
@@ -0,0 +1,40 @@
+#ifndef PHENOTYPE_ANALYSIS_H
```
anyaevostinar, line 1: This file isn't only testing squares, so it should be named something more general, like SymbiontImpact.h

source/sgp_mode/Tasks.h
```diff
@@ -130,14 +131,15 @@ class TaskSet {
 // These are checked top-to-bottom and the reward is given for the first one
 // that matches
 TaskSet DefaultTasks{
-    {"NOT", InputTask{1, [](auto &x) { return ~x[0]; }, 1.0}, false},
```
anyaevostinar, line 133: We should figure out how to have both of these sets of tasks exist in the code without messing each other up. Perhaps this set should be renamed LogicTasks and then the square task should be in a SquareTasks TaskSet and which is used is set by a configuration option?

source/sgp_mode/SGPSymbiont.h
```diff
@@ -36,9 +36,10 @@ class SGPSymbiont : public Symbiont {
 
   void SetHost(emp::Ptr<Organism> host) {
     Symbiont::SetHost(host);
-    cpu.state.used_resources.Delete();
+    //cpu.state.used_resources.Delete();
```
anyaevostinar, line 39: Let's put in a TODO comment so that we don't forget to fix this

source/sgp_mode/Instructions.h
```diff
@@ -96,6 +96,25 @@ INST(IO, {
   *a = next;
   state.input_buf.push(next);
 });
+
+INST(Reuptake, {
+  uint32_t next;
+  float score = state.world->GetTaskSet().CheckTasks(state, *a);
```
anyaevostinar, line 102: Let's put this section of repeated code (repeated from IO) into a small helper method that you can then call from both instructions

source/sgp_mode/Instructions.h
```diff
@@ -96,6 +96,25 @@ INST(IO, {
...
+  if (score != 0.0) {
+    state.host->AddPoints(pow(2, score));
+    if (!state.host->IsHost()) {
```
anyaevostinar, line 105: We should make a note to fix the name of state.host so that it isn't 'host' if it might not always be a host

source/sgp_mode/Instructions.h
```diff
@@ -96,6 +96,25 @@ INST(IO, {
...
+      state.world->sym_points_earned += pow(2, score);
+    }
+  }
+  if(state.internalEnvironment->size() > 0){
+    next = (*state.internalEnvironment)[state.internalEnvironment->size() - 1];
```
anyaevostinar, line 110: It would be good to put in a couple of comments on what is going on here

source/sgp_mode/Instructions.h
```diff
@@ -96,6 +96,25 @@ INST(IO, {
...
+    state.internalEnvironment->pop_back();
+  }
+  else{ next = 0;}
```
anyaevostinar, line 113: I'm not sure we want to give them 0 either since that could be pretty useful to them. Can we just not add to the input buffer if there is nothing in the internal environment?

anyaevostinar (changes requested):
Going in the right direction, a few notes for things to change

Makefile
```diff
@@ -46,7 +46,7 @@ pgg-mode:	source/native/symbulation_pgg.cc
 	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_pgg.cc -o symbulation_pgg
 
 sgp-mode:	source/native/symbulation_sgp.cc
-	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_sgp.cc -o symbulation_sgp
+	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_sgp.cc -o symbulation_sgp -fsanitize=address
```
anyaevostinar, line 49: You should revert this back so that people don't forget to remove it and get stuck with long compile times when it isn't necessary. Remember to git add specific files instead of * so that you don't put in things that you don't mean to be changed

source/native/symbulation_sgp.cc
```diff
@@ -36,8 +36,11 @@ int symbulation_main(int argc, char * argv[])
 
   config.Write(std::cout);
   emp::Random random(config.SEED());
-
-  SGPWorld world(random, &config, DefaultTasks);
+  TaskSet task_set = LogicTasks;
+  if (config.TASK_TYPE() == 0){
+     task_set = SquareTasks;
+  }
```
anyaevostinar, line 42: Make sure to set the DefaultTasks for the other flag value so you don't mess up other people's code

source/native/symbulation_sgp.cc
```diff
@@ -90,13 +93,18 @@ int symbulation_main(int argc, char * argv[])
...
       world.sym_points_donated = 0.0;
       world.sym_points_earned = 0.0;
-      emp::Ptr<SGPHost> host = world.GetFullPop().back().DynamicCast<SGPHost>();
-      if (host->HasSym()){
+      int i = 0;
+      emp::Ptr<SGPHost> host = world.GetFullPop()[i].DynamicCast<SGPHost>();
+      while (!host->HasSym()){//Not always a high number of hosts that have symbionts
+          i++;
+          host = world.GetFullPop()[i].DynamicCast<SGPHost>();
+      }
```
anyaevostinar, line 101: You should probably make sure that you don't go through the whole world by also checking if i equals the size of the population, just in case all the symbionts die out

source/native/symbulation_sgp.cc
```diff
@@ -90,13 +93,18 @@ int symbulation_main(int argc, char * argv[])
...
           emp::Ptr<SGPSymbiont> symbiont = host->GetSymbionts().back().DynamicCast<SGPSymbiont>();
           CheckSymbiont(*host, *symbiont);
-        }  
+        /*else if (host->GetPoints() > 0){
```
anyaevostinar, line 104: Is this code that will be used? If not, let's remove it right away so we don't have dead code lying around

source/ConfigSetup.h
```diff
@@ -76,7 +74,12 @@ EMP_BUILD_CONFIG(SymConfigBase,
 
     GROUP(DTH, "Settings for the Dirty Transmission Hypothesis"),
     VALUE(EFFICIENT_SYM, bool, 0, "Do you want symbionts that also have an efficiency value that evolves"),
+    VALUE(COMPETITION_MODE, bool, 0, "Should a competition between two types be conducted? (Which is specified in the code)"),
 
-
+    GROUP(SGP, "Complex Genomes Settings"),
+    VALUE(CYCLES_PER_UPDATE, size_t, 4, "Number of CPU cycles that organisms run every update"),
+    VALUE(THREAD_COUNT, size_t, 12, "Number of threads used to process organisms in parallel"),
```
anyaevostinar, line 81: You should pull from upstream and go with the upstream default value of 1 for Threat_count

source/sgp_mode/CPU.h
```diff
@@ -27,36 +28,60 @@
...
+      // inst in the library
+      program.resize(100);
+      program[0].op_code = Library::GetOpCode("Global Anchor");
+      program[0].tag = start_tag;
+
+      // io r0
+      // nand r0, r0, r0
+      // io r0
+      // reproduce
+      program[96].op_code = Library::GetOpCode("IO");
```
anyaevostinar, line 78: Pulling from upstream should also turn the IO into SharedIO

source/ConfigSetup.h
anyaevostinar: This needs to be changed back to 1 still

Makefile
```diff
@@ -46,7 +46,7 @@ pgg-mode:	source/native/symbulation_pgg.cc
 	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_pgg.cc -o symbulation_pgg
 
 sgp-mode:	source/native/symbulation_sgp.cc
-	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_sgp.cc -o symbulation_sgp
+	$(CXX_nat) $(CFLAGS_nat) source/native/symbulation_sgp.cc -g -o symbulation_sgp
```
anyaevostinar, line 49: We don't want the g in there either

anyaevostinar (changes requested):
A few more things to fix still

anyaevostinar (approved):
Looks good!

PR #157, opened 2022-07-08
Try to sample a random neighbor in GetNeighborHost and only enumerate all positions if that fails

tolziplohu:
This seems to improve overall performance quite a bit, and the semantics should be exactly the same since it falls back on the old method after three tries with randomly sampled neighbors. In a dense population, it will almost always succeed on the first try and avoid unnecessary computation.

### DIFF
--- a/source/default_mode/SymWorld.h
+++ b/source/default_mode/SymWorld.h
@@ -370,6 +370,14 @@ public:
    * Purpose: To determine the location of a valid occupied neighboring position.
    */
   int GetNeighborHost (size_t i) {
+    // Attempt to use GetRandomNeighborPos first, since it's much faster
+    for (int i = 0; i < 3; i++) {
+      emp::WorldPosition neighbor = GetRandomNeighborPos(i);
+      if (neighbor.IsValid() && pop[neighbor.GetIndex()].Raw())
+        return neighbor.GetIndex();
+    }
+
+    // Then enumerate all occupied neighbors, in case many neighbors are unoccupied
     const emp::vector<size_t> validNeighbors = GetValidNeighborOrgIDs(i);
     if (validNeighbors.empty()) return -1;
     else {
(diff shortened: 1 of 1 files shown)


anyaevostinar:
It looks like a test is failing, did make test pass locally?

anyaevostinar:
The issue is just that you redefine i in the for loop, just change the parameter to "original" or something :)

anyaevostinar:
And please put the position of the parent back in the test since it's good for them to be a bit close

tolziplohu:
> The issue is just that you redefine i in the for loop, just change the parameter to "original" or something :)

Oh, I didn't notice that, I'll fix it!

> And please put the position of the parent back in the test since it's good for them to be a bit close

They're actually close enough though that that test should fail about 1/8 of the time, since if the first host reproduces to a position next to the host that the symbiont is transmitting from, the transmission could go to that host instead of the one the test is testing. If it was position 100 it would be overwritten by the new host anyway, but if the first host is spawned at position 0 there's a 50% chance of the symbiont test failing. I'll try putting it back, but I'm pretty sure it already had a decent chance of failing no matter what.

source/default_mode/SymWorld.h
```diff
@@ -369,8 +369,16 @@ class SymWorld : public emp::World<Organism>{
    *
    * Purpose: To determine the location of a valid occupied neighboring position.
    */
-  int GetNeighborHost (size_t i) {
-    const emp::vector<size_t> validNeighbors = GetValidNeighborOrgIDs(i);
+  int GetNeighborHost (size_t id) {
+    // Attempt to use GetRandomNeighborPos first, since it's much faster
+    for (int i = 0; i < 3; i++) {
+      emp::WorldPosition neighbor = GetRandomNeighborPos(id);
+      if (neighbor.IsValid() && neighbor.GetIndex() < GetSize() && pop[neighbor.GetIndex()].Raw())
```
anyaevostinar, line 376: neighbor.GetIndex() < GetSize() && pop[neighbor.GetIndex()].Raw() can be replaced with IsOccupied(neighbor)

anyaevostinar (changes requested):
Looks good, just a slightly clearer way to check if the neighbor position is occupied and then should be ready to go

anyaevostinar (approved):
Looks great!

PR #161, opened 2022-07-12
Various fixes and additions to complex genomes

tolziplohu:


SymSettings.cfg
```diff
@@ -1,68 +1,80 @@
 ### MAIN ###
```
anyaevostinar, line 1: You should remove SymSettings.cfg from the PR so that we don't have it keep changing

PR #172, opened 2022-08-05
CheckSymbiont() and GetDominantInfo()

tolziplohu:


source/sgp_mode/CPU.h
```diff
@@ -77,7 +87,8 @@ class CPU {
    * Purpose: Mutates the genome code stored in the CPU.
    */
   void Mutate() {
-    program.ApplyPointMutations(0.03);
+    program.ApplyPointMutations(state.world->GetConfig()->MUTATION_SIZE() *
```
anyaevostinar, line 90: This should probably be attached to MUTATION_RATE instead of size, since size doesn't really make sense when we're talking about instructions instead of numbers

anyaevostinar (approved):

source/sgp_mode/CPU.h
tolziplohu: Yeah, I was thinking about that but the main reason I used MUTATION_SIZE is because the default value is pretty reasonable, whereas MUTATION_RATE defaults to 1; also, the description of MUTATION_RATE doesn't really fit because it describes it as the chance that any mutation occurs at all, whereas this is more like how much to mutate and thus closer to MUTATION_SIZE.

PR #173, opened 2022-08-08
Complex genomes

anyaevostinar:
Getting AnalysisTools into my dev branch


.vscode/configurationCache.log
```diff
@@ -0,0 +1 @@
+{"buildTargets":[],"launchTargets":[],"customConfigurationProvider":{"workspaceBrowse":{"browsePath":[],"compilerArgs":[]},"fileIndex":[]}}
```
anyaevostinar, line 1: You should add .vscode to your gitignore so you don't have all these vscode files

source/ConfigSetup.h
```diff
@@ -78,7 +78,7 @@ EMP_BUILD_CONFIG(SymConfigBase,
 
     GROUP(SGP, "Complex Genomes Settings"),
     VALUE(CYCLES_PER_UPDATE, size_t, 4, "Number of CPU cycles that organisms run every update"),
-    VALUE(THREAD_COUNT, size_t, 12, "Number of threads used to process organisms in parallel"),
+    VALUE(THREAD_COUNT, size_t, 4, "Number of threads used to process organisms in parallel"),
```
anyaevostinar, line 81: What is this change from?

source/catch/main.cc
```diff
@@ -38,5 +40,5 @@
 #include "../test/integration_test/endosymbiosis/res_distribute.test.cc"
 #include "../test/integration_test/dirty_transmission/hz_mut_rate.test.cc"
 
-//#include "../PGGendtoend.test.cc"
-//#include "../test/end_to_end.test.cc"
+// #include "../PGGendtoend.test.cc"
```
anyaevostinar, line 43: Please remove these since we're trying to keep them out finally

source/native/symbulation_sgp.cc
```diff
@@ -26,9 +26,9 @@ int symbulation_main(int argc, char * argv[])
 
   TaskSet task_set = LogicTasks;
   if (config.TASK_TYPE() == 0){
-     task_set = SquareTasks;
+     task_set = LogicTasks;
```
anyaevostinar, line 29: This appears to be reversed from what we have in my repo

source/test/sgp_mode_test/SGPModularity.test.cc
```diff
@@ -0,0 +1,389 @@
...
+                                            0, 0, 0, 0, 0, 0, 0, 0, 0, 1};
+    emp::vector<int> needed_code_sites_e = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                                            0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+    emp::vector<emp::vector<int>> needed_code_sites = {
+        needed_code_sites_a, needed_code_sites_b, needed_code_sites_c,
+        needed_code_sites_d, needed_code_sites_e};
+
+    emp::vector<int> calc_useful_starts = GetUsefulStarts(needed_code_sites);
+    emp::vector<int> true_useful_starts = {0, 6, needed_code_sites_c.size() - 5,
```
anyaevostinar, line 40: This line is causing a 'narrowing conversion' error on my compiler and GitHubs, so you should fix it, specifically the problem is the 'needed_code_sites_c.size() -5' "error: non-constant-expression cannot be narrowed from type
      'unsigned long' to 'int' in initializer list"

.vscode/configurationCache.log
dorchens: added to my git ignore

source/ConfigSetup.h
dorchens: My personal computing was getting very loud when I was running it on the twelve so I lowered it to 4 for my own sake

SymSettings.cfg
```diff
@@ -12,7 +12,7 @@ set GRID_Y 100                    # Height of world, just multiplied by width to
 set POP_SIZE -1                   # Starting size of the host population, -1 for full starting population
 set SYM_LIMIT 1                   # Number of symbiont allowed to infect a single host
 set START_MOI 1                   # Ratio of symbionts to hosts that experiment should start with
-set UPDATES 1001                  # Number of updates to run before quitting
+set UPDATES 1001               # Number of updates to run before quitting
```
anyaevostinar, line 15: Looks like you accidentally deleted some space so this line doesn't match the rest

source/catch/main.cc
dorchens: Pretty sure they should be gone now

source/native/symbulation_sgp.cc
dorchens: Weird yeah the switch shouldn't make a difference so I'll just change it on my fork and accept this change

dorchens (approved):
Everything should be fixed now! No more non clang-d errors and the code set up looks like yours now in the settings

PR #175, opened 2022-08-10
Top N dominant hosts, and set SGP seed based on world seed

tolziplohu:


source/native/symbulation_sgp.cc
```diff
@@ -36,21 +36,21 @@ int symbulation_main(int argc, char *argv[]) {
   world.CreateDataFiles();
   world.RunExperiment();
 
-  std::pair<emp::Ptr<Organism>, size_t> info = world.GetDominantInfo();
-  std::cout << "Dominant count: " << info.second << std::endl;
+  emp::vector<std::pair<emp::Ptr<Organism>, size_t>> info =
```
anyaevostinar, line 39: Since "info" is used quite a bit, let's give it a clearer name, such as "dominant_organisms"?

anyaevostinar (changes requested):
Just should improve one variable name, otherwise looks good

anyaevostinar (approved):

PR #177, opened 2022-08-11
Complex genomes

dorchens:
New pull request for modularity analysis, and analysis tools code


.gitignore
```diff
@@ -15,3 +15,5 @@ symbulation.test
 .Rproj.user
 *.gcno
 *.gcda
+signalgp-lite
```
anyaevostinar, line 18: We don't want signalgp-lite ignored, so please remove it. If you need it in your local .gitignore, you should just make sure not to add and push this file

source/sgp_mode/AnalysisTools.h
```diff
@@ -103,7 +104,7 @@ emp::vector<int> GetNecessaryInstructions(CPU org_cpu, size_t test_task_id) {
 
   if (can_do_task) {
 
-    for (int k = 0; k <= control_program.size() - 1; k++) {
+    for (int k = 0; k <= (int)control_program.size() - 1; k++) {
```
anyaevostinar, line 107: it's better to use 'size_t' as the type within forloops

source/sgp_mode/CPU.h
```diff
@@ -89,7 +90,14 @@ class CPU {
...
     cpu.InitializeAnchors(program);
   }
-
+  
+  /**
+   * Input: None
+   *
+   * Output: Returns the CPU's program
+   *
+   * Purpose: To Get the Program of an Prganism from its CPU
```
anyaevostinar, line 99: typo "Prganism"

source/sgp_mode/ModularityAnalysis.h
```diff
@@ -46,8 +46,8 @@
 emp::vector<int> GetUsefulStarts(emp::vector<emp::vector<int>> task_programs) {
   emp::vector<int> list_of_starts = {};
 
-  for (int y = 0; y < task_programs.size(); y++) {
-    for (int e = 0; e <= task_programs[y].size() - 1; e++) {
+  for (int y = 0; y < (int)task_programs.size(); y++) {
```
anyaevostinar, line 49: same for these for loops, would be better as 'size_t' and that will probably remove the need to cast to an int

source/sgp_mode/ModularityAnalysis.h
```diff
@@ -193,7 +192,7 @@ double GetPModularity(emp::vector<emp::vector<int>> task_programs) {
 
 /**
  *
- * Input:Takes in the number of tasks in the taskset and a pointer to a host
+ * Input:Takes in the number of tasks in the taskset and a CPU reference
```
anyaevostinar, line 195: Looks like it only takes a CPU reference now

source/sgp_mode/ModularityAnalysis.h
```diff
@@ -263,7 +262,7 @@ double GetFModularity(emp::vector<emp::vector<int>> task_programs) {
...
```
anyaevostinar, line 265: Ditto here, only takes a CPU

source/test/sgp_mode_test/AnalysisTools.test.cc
```diff
@@ -38,75 +35,90 @@ TEST_CASE("ReturnTaskDone", "[sgp]") {
...
+      emp::Ptr<SGPHost> test_sample =
+          emp::NewPtr<SGPHost>(&random, &world, &config);
+
+      test_map = GetReducedProgramRepresentations(test_sample->GetCPU());
+      emp::vector<emp::vector<int>> expected_map = {
+          expected_vector, {-1}, {-1}, {-1}, {-1}, {-1}, {-1}, {-1}, {-1}};
+      for (int guide = 0; guide < (int)test_map.size(); guide++) {
+        THEN("There are 4 needed code sites necessary to perform the Not-task "
+             "and the genome cannot perform any other tasks") {
+          REQUIRE(test_map == expected_map);
```
anyaevostinar, line 119: You are looping the size of the map, but just comparing the maps directly and not using the 'guide' variable, so I don't think you need the for loop

anyaevostinar (changes requested):
A few things to change

anyaevostinar (approved):

PR #181, opened 2022-08-16
Use inheritance for InputTask and OutputTask, & guaranteed vertical transmission in no mut updates

tolziplohu:


source/sgp_mode/CPU.h
```diff
@@ -45,19 +45,38 @@ class CPU {
...
     cpu.Reset();
-    cpu.InitializeAnchors(program);
     state = CPUState(state.host, state.world);
+    InitializeAnchors();
     state.self_completed.resize(state.world->GetTaskSet().NumTasks());
     state.shared_completed->resize(state.world->GetTaskSet().NumTasks());
   }
-    
+
+  void InitializeAnchors() {
```
anyaevostinar, line 61: Documentation for this?

anyaevostinar (approved):

source/sgp_mode/CPU.h
tolziplohu: Good idea, I've added some documentation (and made this method private) in my fork now so that can be in the next PR.

PR #184, opened 2022-08-26
Fix bugs related to calculation of modularity and diversity

tolziplohu:
The `BitArray` change was originally for performance (we were allocating a length-100 vector of integers which were all 0 or 1, it could just be two 64-bit integers instead) but the segfault was also fixed incidentally by that change, it could also have been done without changing all the types. The other change with the reuptake instruction stops organisms from simply outputting the same value over and over again and getting a reward every time, which they had been doing before and which makes calculating phenotypes essentially impossible since the task completions are just relying on certain values being in their registers.

### DIFF
--- a/source/sgp_mode/CPU.h
+++ b/source/sgp_mode/CPU.h
@@ -109,18 +109,6 @@ public:
 
     sgpl::execute_cpu_n_cycles<Spec>(n_cycles, cpu, program, state);
   }
-  /**
-   * Input: None
-   *
-   * Output: The SignalGPLite cpu. 
-   *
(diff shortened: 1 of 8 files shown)


source/sgp_mode/CPU.h
```diff
@@ -109,18 +109,6 @@ class CPU {
 
     sgpl::execute_cpu_n_cycles<Spec>(n_cycles, cpu, program, state);
   }
-  /**
```
anyaevostinar, line 112: Did you mean to delete this method?

PR #189, opened 2022-10-11
Adding Random and Normalized Initialization of the symbionts' donation rate in PGG mode

Apm0816:
Added 6 configuration settings, two are booleans that show whether the user has chosen for the donation rate to be randomly selected and to be selected using a normal solution. Two settings are used to set a min and max for a range for the initialization, and the final two are for the mean and standard deviation for a normal distribution. These settings are used in PGGWorldSetup.cc where the donation rate for each symbiont is selected.


emilydolson:
@anyaevostinar I've looked through this and don't see any issues, so feel free to merge if it looks good to you!

anyaevostinar (approved):
Looks good!

PR #191, opened 2022-10-18
Adding randomized and normalized initialization of the PGG donation rate

Apm0816:
Added support for randomized and normalized initialization of the PGG donation rate. This was done by adding new settings for the PGG game in the SymSettings.cfg file.


emilydolson (approved):
This PR just adds the new configs to SymSettings.cfg (we forgot to in #189)

PR #194, opened 2022-11-02
Intergration tests

K-Johnson-Horrigan:
Integration tests checking behavior:
- dirty transmission: symbionts are parasitic at low vertical transmission rates, mutualistic at high vertical transmission rates, and at intermediate vertical transmission rates they are more mutualistic when horizontal transmission mutation rate is higher.
- vertical transmission: starting distribution of behaviour (mixed, neutral, parasitic, mutualistic), vertical transmission rates, and spatial structure affect the evolution of symbiotic behavior.
- multi-infection: when there can be many symbionts per host they are less likely to evolve mutualistic behavior in some conditions.


anyaevostinar (approved):
Looks great!

Issue #190, opened 2022-10-18
Systematics manager crashes with seg fault at end


emilydolson:
When phylogeny tracking is on, Symbulation crashes with a seg fault during the final clean-up. This is due to some messy stuff in the sequencing of the destructors. Basically, what happens is the following:

1. The SymWorld destructor is called.  
2. In the SymWorld destructor, we delete the symbiont systematics manager.
3. Once the SymWorld destructor is done, we automatically go into the destructor for its base class (emp::world<Organism>).
4. The base class destructor goes through the population and deletes each host.
5. The host destructor goes through the host's symbionts and deletes each of them.
6. The symbiont destructor tries to grab the symbiont systematics manager and notify it that this symbiont should be removed.
7. The systematics manager has already been deleted, so this produces a seg fault.

I'm not sure what the best way to fix this is. I came up with some possible solutions, but they're all kind of unpleasant:
- The symbiont systematics manager could get moved to some sort of scope that lives outside the SymWorld
- We could add an extra check in the symbiont destructor that confirms that the systematics manager has not been deleted yet (it's annoying to have to do that check every time a symbiont is deleted just to avoid this one problem at the very end of the run, though)
- We could move the code that notifies the systematics manager that the symbiont has died to `SetDead`. Is that reliably called every time a Symbiont dies? If so, this is probably the best option.

emilydolson:
Fixed by #195

PR #195, opened 2022-11-04
delete hosts before deleting sym systematic

K-Johnson-Horrigan:


emilydolson (commented):
Oh, wow, this is a way better solution to #190 than any of the ones I came up with! Looks great! 

If it's not too much trouble, adding a small test that enforces that the phylogeny gets destructed properly and doesn't seg-fault might be worthwhile.

anyaevostinar:
@K-Johnson-Horrigan Would it be possible to do the small test that Emily mentioned? Probably just creating the world with phylogeny on and then deleting it and making sure it doesn't crash would work?

anyaevostinar (changes requested):
Let's try to do the test that is mentioned in the comment thread to catch this kind of error in the future

anyaevostinar (approved):
Great job!

PR #197, opened 2022-11-05
Sgp organism counts data file

K-Johnson-Horrigan:
Adds a data file that stores total host and symbiont counts per update; if free living symbionts are permitted, they are also counted and written to the data file.

### DIFF
--- a/source/sgp_mode/SGPDataNodes.h
+++ b/source/sgp_mode/SGPDataNodes.h
@@ -9,12 +9,13 @@
 void SGPWorld::CreateDataFiles() {
   std::string file_ending =
       "_SEED" + std::to_string(my_config->SEED()) + ".data";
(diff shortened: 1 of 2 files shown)


source/sgp_mode/SGPDataNodes.h
```diff
@@ -11,6 +11,10 @@ void SGPWorld::CreateDataFiles() {
       "_SEED" + std::to_string(my_config->SEED()) + ".data";
   // SGP mode doesn't need int val files, and they have significant performance
   // overhead. Only the transmission file needs to be created for this mode.
+  SetUpOrgCountFile(my_config->FILE_PATH() + "OrganismCounts" +
```
anyaevostinar, line 14: Should move this down so that the previous comment is still above the correct line

anyaevostinar (approved):
Looks great!

PR #198, opened 2022-11-11
Organism refactor with multiple inheritance

tolziplohu:
I don't know if we actually want to go through with all of this, I'm basically trying to take it as far as I can and then we can figure out how much we want to keep. This PR also adds more `cc` files, but they're still just `#include`d instead of actually being compiled separately because Empirical doesn't support that. Unfortunately, that means they have to be included in each entry point file - currently only default mode is working, I'll update SGP mode next.


source/pgg_mode/PGGHost.h
```diff
@@ -111,7 +113,11 @@ class PGGHost: public Host {
     Host::DistribResources(resources);
 
     for(size_t i=0; i < syms.size(); i++){
-      double hostPool = syms[i]->ProcessPool();
+      emp::Ptr<PGGSymbiont> sym = syms[i].DynamicCast<PGGSymbiont>();
+      if (sym == nullptr) {
+        throw "Non-PGG symbiont in PGG host";
+      }
+      double hostPool = sym->ProcessPool();
```
tolziplohu, line 120: This is the worst part of this change in my opinion, unfortunately I haven't really found any way to get around it without either adding stubs to BaseSymbiont (so the "Organism has every method needed by any mode" approach used before this PR) or using far too many templates.

tolziplohu:
After an unreasonable amount of time spent debugging this, it looks like Empirical's pointer tracking infrastructure doesn't actually support inheritance correctly. It's only appearing now that the inheritance is more complex with the multiple inheritance changes, but it's not actually specific to multiple inheritance - it assumes that the address of an object doesn't change when it's casted to a base class, but that's not actually true in general. Specifically, we create a pointer to `Host`, and Empirical notes down its address as a valid pointer; then we use it as a pointer to `Organism`, and the compiler cuts off the `Host` vtables and moves the pointer forward a few bytes. When we try to delete it, the compiler knows how to delete the correct object, but to Empirical it looks like we're deleting an object that was never allocated, since the address is different.

I'm not sure what the best approach to get around this is, but the easiest option would be to just give up on Empirical's pointer tracking - remove `-DEMP_TRACK_MEM` from the debug mode flags. In my experience AddressSanitizer works significantly better anyway, but if you want to keep Empirical's pointer tracking, we can try to find another solution.

PR #200, opened 2022-11-12
SGP ectosymbiosis

K-Johnson-Horrigan:
- fixing a memory issue Anya pointed out from PR196 
- implementing basic ectosymbiosis; at the start of "steal" and "donate," free-living symbionts with parallel hosts who are open to ectosymbiosis "link" to those hosts, and at the end of such instructions they "unlink," so that other free-living functions (e.g. "infect") can proceed unimpeded.

### DIFF
--- a/source/sgp_mode/SGPSymbiont.h
+++ b/source/sgp_mode/SGPSymbiont.h
@@ -85,13 +85,16 @@ public:
   void SetHost(emp::Ptr<Organism> host) {
     Symbiont::SetHost(host);
     if (my_host) {
+      if (cpu.state.used_resources) { cpu.state.used_resources.Delete(); }
+      if (cpu.state.shared_available_dependencies) {
+        cpu.state.shared_available_dependencies.Delete(); }
+      if (cpu.state.internalEnvironment) { cpu.state.internalEnvironment.Delete(); }
+
       cpu.state.used_resources =
         host.DynamicCast<SGPHost>()->GetCPU().state.used_resources;
-    cpu.state.shared_available_dependencies =
-        host.DynamicCast<SGPHost>()
-            ->GetCPU()
-            .state.shared_available_dependencies;
-    cpu.state.internalEnvironment =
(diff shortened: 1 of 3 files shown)


source/sgp_mode/SGPSymbiont.h
```diff
@@ -6,6 +6,7 @@
 #include "SGPHost.h"
 #include "SGPWorld.h"
 #include "emp/Evolve/World_structure.hpp"
+#include "SGPHost.h"
```
anyaevostinar, line 9: Duplicate include line

source/sgp_mode/SGPSymbiont.h
```diff
@@ -85,13 +86,16 @@ class SGPSymbiont : public Symbiont {
   void SetHost(emp::Ptr<Organism> host) {
     Symbiont::SetHost(host);
     if (my_host) {
+      if (cpu.state.used_resources) { cpu.state.used_resources.Delete(); }
+      if (cpu.state.shared_available_dependencies) {
+        cpu.state.shared_available_dependencies.Delete(); }
```
anyaevostinar, line 91: Are we certain that these are pointers that the Symbiont owns? Aren't some of these potentially since being used by the old host? Or should be cleaned up by the old host upon death?

anyaevostinar (commented):
One concern about memory management, otherwise looks good

anyaevostinar (changes requested):
Also a duplicate line to delete

source/sgp_mode/SGPSymbiont.h
anyaevostinar: I think the plan is to not delete these right? Leave them to the host to delete?

anyaevostinar (changes requested):
Just need to not delete the host's pointers for future scenarios where symbionts jump out of hosts

anyaevostinar (approved):

PR #201, opened 2022-12-01
SGP free living symbionts different repro threshold

K-Johnson-Horrigan:
Free living syms can have a different reproductive threshold (# resources they need in order to reproduce) than hosted symbionts.


anyaevostinar (approved):
Looks good

PR #202, opened 2022-12-01
World::Setup changes for Complex Genomes without the class structure changes

tolziplohu:
This way it's a lot easier to look at the class structure changes on their own instead of it all being in the same pull request.


anyaevostinar (approved):
These includes seem harmless, so I'm letting them through, if they are actually not needed, we can remove them later

PR #205, opened 2022-12-09
Config inheritance

K-Johnson-Horrigan:
Non-default (lysis, efficient, and PGG) modes have their own ConfigSetup files which inherit from base and include mode-specific settings.


source/efficient_mode/EfficientHost.h
```diff
@@ -20,17 +20,25 @@ class EfficientHost: public Host {
...
 public:
   /**
    * The constructor for efficient host
    */
-  EfficientHost(emp::Ptr<emp::Random> _random, emp::Ptr<EfficientWorld> _world, emp::Ptr<SymConfigBase> _config,
+  EfficientHost(emp::Ptr<emp::Random> _random, emp::Ptr<EfficientWorld> _world, emp::Ptr<SymConfigEfficient> _config,
   double _intval =0.0, emp::vector<emp::Ptr<Organism>> _syms = {},
   emp::vector<emp::Ptr<Organism>> _repro_syms = {},
   double _points = 0.0, double _efficient = 0.1) :
   Host(_random, _world, _config, _intval, _syms, _repro_syms, _points) {
```
anyaevostinar, line 38: I'm worried that this will cause there to be two config objects, one used by the super class methods and one the subclass. Are you sure that you need to declare the type differently or is it real inheritance where we could refer to the config object by the base name still?

source/ConfigSetup.h
```diff
@@ -31,6 +31,7 @@ EMP_BUILD_CONFIG(SymConfigBase,
     VALUE(PHYLOGENY, bool, 0, "Should the world keep track of host and symbiont phylogenies? (0 for no, 1 for yes)"),
     VALUE(NUM_PHYLO_BINS, size_t, 5, "How many bins should organisms be sepeated into if phylogeny is on?"),
     VALUE(NO_MUT_UPDATES, int, 0, "How many updates should be run after the end of UPDATES with all mutation turned off?"),
+    VALUE(SYM_EXCLUDE, bool, 0, "Do symbionts have a decreasing chance of getting into the host the more symbionts are already infecting it? (0 for no, 1 for yes)"),
```
anyaevostinar, line 34: Let's include an alias so that old config files will still work for the new name, should just go right after the description string

anyaevostinar: You could also then test it by not changing the tests :)

source/default_mode/Host.h
```diff
@@ -461,8 +461,7 @@ class Host: public Organism {
    * where n is the number of existing phage.
    */
   bool SymAllowedIn(){
-    bool do_phage_exclusion = my_config->PHAGE_EXCLUDE();
-    if(!do_phage_exclusion){
+    if(!my_config->SYM_EXCLUDE()){
```
anyaevostinar, line 464: Even though you could not change this (after the alias is in), I think it's good to keep this name change so that it isn't confusing in the code why we are talking about phage

source/efficient_mode/EfficientConfigSetup.h
```diff
@@ -0,0 +1,12 @@
+#ifndef EFFICIENT_CONFIG_H
+#define EFFICIENT_CONFIG_H
+#include "../../Empirical/include/emp/config/config.hpp"
+#include "../ConfigSetup.h"
+
+EMP_EXTEND_CONFIG(SymConfigEfficient, SymConfigBase,
+  GROUP(DTH, "Settings for the Dirty Transmission Hypothesis"),
+  VALUE(EFFICIENT_SYM, bool, 0, "Do you want symbionts that also have an efficiency value that evolves"),
+  VALUE(COMPETITION_MODE, bool, 0, "Should a competition between two types be conducted? (Which is specified in the code)")
```
anyaevostinar, line 9: Competition mode code all got cut out right? So we could remove this configuration option probably

source/efficient_mode/EfficientHost.h
```diff
@@ -20,17 +20,25 @@ class EfficientHost: public Host {
     *
   */
   emp::Ptr<EfficientWorld> my_world = NULL;
+
+  /**
+    *
+    * Purpose: Represents the configuration settings for a particular run.
```
anyaevostinar, line 26: Let's put in a note here to be more verbose since this can be confusing. "Note: holds all configuration settings and points to same configuration object as my_config, but with the correct subtype."

source/efficient_mode/EfficientSymbiont.h
```diff
@@ -49,22 +49,30 @@ class EfficientSymbiont: public Symbiont {
     *
   */
   emp::Ptr<EfficientWorld> my_world = NULL;
+
+  /**
+    *
+    * Purpose: Represents the configuration settings for a particular run.
```
anyaevostinar, line 55: Same as previous update

source/efficient_mode/EfficientWorld.h
```diff
@@ -3,17 +3,30 @@
 
 #include "../default_mode/SymWorld.h"
 #include "../default_mode/DataNodes.h"
+#include "EfficientConfigSetup.h"
 
 class EfficientWorld : public SymWorld {
 private:
+  /**
+    *
+    * Purpose: Represents the configuration settings for a particular run.
```
anyaevostinar, line 12: Ditto as previous

source/efficient_mode/EfficientWorld.h
```diff
@@ -3,17 +3,30 @@
...
   /**
     *
     * Purpose: Data node tracking the average efficiency of efficient symbionts.
     *
   */
   emp::Ptr<emp::DataMonitor<double>> data_node_efficiency;
 public:
-  using SymWorld::SymWorld;
+  /**
+   * The constructor for PGGWorld
```
anyaevostinar, line 25: Need to fix documentation, also should include full documentation of what the parameters are

source/lysis_mode/LysisWorld.h
```diff
@@ -3,9 +3,18 @@
 
 #include "../default_mode/SymWorld.h"
 #include "../default_mode/DataNodes.h"
+#include "LysisConfigSetup.h"
 
 class LysisWorld : public SymWorld {
 private:
+
+  /**
+    *
```
anyaevostinar, line 12: Ditto

source/lysis_mode/LysisWorld.h
```diff
@@ -19,7 +28,12 @@ class LysisWorld : public SymWorld {
   emp::Ptr<emp::DataMonitor<int>> data_node_cfu;
 
 public:
-  using SymWorld::SymWorld;
+  /**
+   * The constructor for LysisWorld
```
anyaevostinar, line 32: Needs Input info

source/pgg_mode/PGGHost.h
```diff
@@ -22,11 +22,21 @@ class PGGHost: public Host {
   */
   emp::Ptr<PGGWorld> my_world = NULL;
 
+  /**
+    *
+    * Purpose: Represents the configuration settings for a particular run.
```
anyaevostinar, line 27: Ditto

source/pgg_mode/PGGSymbiont.h
```diff
@@ -21,9 +21,18 @@ class PGGSymbiont: public Symbiont {
   */
   emp::Ptr<PGGWorld> my_world = NULL;
 
+  /**
+    *
```
anyaevostinar, line 25: Ditto

source/pgg_mode/PGGWorld.h
```diff
@@ -3,17 +3,30 @@
 
 #include "../default_mode/SymWorld.h"
 #include "../default_mode/DataNodes.h"
+#include "PGGConfigSetup.h"
 
 class PGGWorld : public SymWorld {
 private:
+  /**
+    *
+    * Purpose: Represents the configuration settings for a particular run.
```
anyaevostinar, line 12: Ditto

source/pgg_mode/PGGWorld.h
```diff
@@ -3,17 +3,30 @@
...
   /**
     *
     * Purpose: Data node tracking the pgg donation rate.
     *
   */
   emp::Ptr<emp::DataMonitor<double,emp::data::Histogram>> data_node_PGG;
 public:
-  using SymWorld::SymWorld;
+  /**
+   * The constructor for PGGWorld
```
anyaevostinar, line 25: Need inputs listed

source/test/lysis_mode_test/Bacterium.test.cc
```diff
@@ -199,60 +199,9 @@ TEST_CASE("Bacterium Process", "[lysis]"){
     }
 }
 
-TEST_CASE("Phage Exclude", "[lysis]") {
```
anyaevostinar, line 202: Could just leave this test to test the aliasing

anyaevostinar (changes requested):
Just some documentation to clarify and an alias to add

anyaevostinar (approved):

PR #208, opened 2024-03-25
Update to latest version of Empirical

emilydolson:
When convenient, these changes should make Symbulation work with the latest changes to Empirical


emilydolson:
Oh, I see. This breaks compatabilility with the version of emscripten used to compile the web tests. I'll try to take a look at that soon.

anyaevostinar (approved):

PR #210, opened 2024-04-18
Copilot

bennetthilberg:
Copilot is an add-on for making tutorials and other guidance with react/javascript that includes a premade tutorial.


anyaevostinar (approved):
Looks great!

PR #218, opened 2024-10-02
default ousting config option + test

K-Johnson-Horrigan:


source/default_mode/Host.h
```diff
@@ -439,7 +439,13 @@ class Host: public Organism {
    * Purpose: To add a symbionts to a host's symbionts
    */
   int AddSymbiont(emp::Ptr<Organism> _in) {
-    if((int)syms.size() < my_config->SYM_LIMIT() && SymAllowedIn()){
+    bool allowed_in = SymAllowedIn();
+    if (my_config->OUSTING() && allowed_in && (int)syms.size() == my_config->SYM_LIMIT()) {
+      emp::Ptr<Organism> old_sym = syms.back();
```
anyaevostinar, line 444: Random sym would be better than last

anyaevostinar: Check if only one sym, if more than one, then figure out a random one to oust

anyaevostinar: Swap the randomly chosen one to die with the last one for efficiency

anyaevostinar: Make sure to return the location where the new sym actually ended up

source/default_mode/Host.h
```diff
@@ -439,7 +439,17 @@ class Host: public Organism {
    * Purpose: To add a symbionts to a host's symbionts
    */
   int AddSymbiont(emp::Ptr<Organism> _in) {
-    if((int)syms.size() < my_config->SYM_LIMIT() && SymAllowedIn()){
+    bool allowed_in = SymAllowedIn();
+    if (my_config->OUSTING() && allowed_in && (int)syms.size() == my_config->SYM_LIMIT()) {
+      int new_sym_pos = (syms.size() > 1) ? random->GetInt(syms.size()) : syms.size()-1;
```
anyaevostinar, line 444: I'm unfairly biased against ternary operators and have a hard time following them, can this be either just a normal if or have a comment explaining what it is doing?

anyaevostinar (approved):

PR #224, opened 2024-11-01
Stress Hosts

K-Johnson-Horrigan:
Additions:
- Stress host subclass
- Vertical transmission task matching as config option
- SGP organism type (SGP, Health, Stress) as config option


source/native/symbulation_parasite.cc
```diff
@@ -28,6 +28,20 @@ int symbulation_main(int argc, char *argv[]) {
   SymConfigSGP config;
   CheckConfigFile(config, argc, argv);
 
+  // stress hard-coded transmission modes
+  if (config.ORGANISM_TYPE() == 2) {
```
amlalejini, line 32: We can create an organism type enumerated class to check against instead of directly checking against the numbers (to improve readability).

amlalejini: (same for the checks below)

source/sgp_mode/SGPWorld.h
```diff
@@ -164,7 +166,8 @@ class SGPWorld : public SymWorld {
 
   
   emp::WorldPosition SymDoBirth(emp::Ptr<Organism> sym_baby, emp::WorldPosition parent_pos) override;
-  int GetNeighborHost (size_t id, emp::Ptr<emp::BitSet<64>>);
```
amlalejini, line 167: Add a constexpr variable somewhere for this bitset size? Probably want to define it in whatever location is dictating the choice of value.

amlalejini (approved):

PR #250, opened 2025-07-01
Added Steal and Donate Health host with config

EliasH-E:
This allows symbionts of a Health host to gain "donate" and "steal" instructions. Currently, both donate and steal are both available, in order to turn off one of these, comment it out in the library section of GenomeLibrary.h.

### DIFF
--- a/source/Organism.h
+++ b/source/Organism.h
@@ -75,6 +75,7 @@ class Organism {
     std::cout << "MakeNew called from Organism" << std::endl;
     throw "Organism method called!";
   }
+  
 
   //Symbiont functions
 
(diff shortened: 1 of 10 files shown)


source/Organism.h
```diff
@@ -75,6 +75,7 @@ class Organism {
     std::cout << "MakeNew called from Organism" << std::endl;
     throw "Organism method called!";
   }
+
```
anyaevostinar, line 78: remove extra  line

source/sgp_mode/HealthHost.h
```diff
@@ -62,9 +64,26 @@ class HealthHost : public SGPHost {
         int host_cycle = 1;
         int sym_cycle = 0;
         if (HasSym()) {
-          if (sgp_config->STRESS_TYPE() == MUTUALIST) {
+
```
anyaevostinar, line 67: extra line

PR #252, opened 2025-07-07
Addition of Tests and reorganizing of previous tests

EliasH-E:
The 3 tests files I have created in the past I folded into other test files so it matched the practice of tying tests with scripts. Then I also created new tests for SGPWorld.test.cc which focuses on the setup function. The creation of this test involved the creation of new override name functions for HealthHost.h, SGPHost.h and StressHost.h


source/sgp_mode/GenomeLibrary.h
```diff
@@ -256,6 +256,16 @@ class ProgramBuilder : emp::vector<sgpl::Instruction<Spec>> {
     Add("Nand", 0, 0, 0);
     Add("SharedIO");
   }
+
+  void AddSteal(){
+    
+    Add("Steal");
+    //std::cout << "Adding Steal" << std::endl;
```
anyaevostinar, line 263: remove comment
Also why are these needed? Why can't you just call "Add("Steal") in the create programs?

source/sgp_mode/HealthHost.h
```diff
@@ -7,6 +7,8 @@ class HealthHost : public SGPHost {
 
     public:
 
+    int cycles_given = 0;
```
anyaevostinar, line 10: There is a lot of code repeated between the two PRs it looks like, which will be a problem. Each PR should be totally independent code, so if they aren't actually able to be independent, they need to be back together as one

anyaevostinar (changes requested):
It looks like much of this code is repeated in the other PR so this one needs to wait or they need to be combined again

source/test/sgp_mode_test/SGPDataNodes.test.cc
```diff
@@ -0,0 +1,210 @@
...
+        REQUIRE(str4 == "0,0,0,0,0");
+      }
+      std::getline(file4, str4);
+      THEN("The third should be at 100 updates with 7 attempts and 1 success"){
+        REQUIRE(str4 == "100,7,1,0,0");
+      }
+  }
+
+
+  //std::filesystem::remove("OrganismCountsDataTest_SEED2.data");
```
anyaevostinar, line 204: Why are these commented out? Are they needed or no? If not needed, remove the lines, otherwise uncomment them

source/test/sgp_mode_test/SGPWorld.test.cc
```diff
@@ -74,6 +74,66 @@ TEST_CASE("Baseline function", "[sgp]") {
...
+
+
+  //world.SetupHosts requires a pointer for the number of hosts in the world
+  unsigned long setupCount = 1;
+  WHEN("ORGANISM_TYPE Config is set to SGP hosts"){
+    config.ORGANISM_TYPE(0);
+    SGPWorld world(random, &config, LogicTasks);
+    world.SetupHosts(&setupCount);
+    THEN("The world contains a SGPHost"){
+    emp::Ptr<SGPHost> host =  (world.GetOrgPtr(0)).Cast<SGPHost>();
```
anyaevostinar, line 97: It seems like casting it shouldn't be necessary to check its type?

anyaevostinar (changes requested):
Looking really good, just a couple more things

anyaevostinar (approved):

PR #253, opened 2025-07-08
Added tests for GenomeLibrary and SGPSymbiont in SGP mode

TahaKhalili:


source/test/sgp_mode_test/SGPSymbiont.test.cc
```diff
@@ -51,15 +56,63 @@ TEST_CASE("SGPSymbiont Reproduce", "[sgp]") {
...
+      for (unsigned int i = 1; i < CPU_BITSET_LENGTH; i++)
+      {
         REQUIRE(sym_baby->GetCPU().state.task_toward_partner[i] == 0);
         REQUIRE(sym_baby->GetCPU().state.task_from_partner[i] == 0);
       }
     }
     sym_baby.Delete();
   }
+
+  SECTION("Symbiont comparison operators")
```
anyaevostinar, line 73: I think these should be in their own test case instead of just a separate section

source/test/sgp_mode_test/SGPSymbiont.test.cc
```diff
@@ -51,15 +56,63 @@ TEST_CASE("SGPSymbiont Reproduce", "[sgp]") {
...
+      REQUIRE(*sym_parent == *clone1);
+      REQUIRE(*clone1 == *clone2);
+    }
+
+    THEN("Symbionts with different genomes are not equal")
+    {
+      REQUIRE_FALSE(*sym_parent == *different);
+    }
+
+    THEN("operator< reflects program ordering")
```
anyaevostinar, line 90: Also it's own test

anyaevostinar (changes requested):
Great tests, just a bit of formatting to change/clean up

source/test/sgp_mode_test/SGPSymbiont.test.cc
```diff
@@ -56,39 +50,33 @@ TEST_CASE("SGPSymbiont Reproduce", "[sgp]")
...
         REQUIRE(sym_baby->GetCPU().state.task_toward_partner[i] == 0);
         REQUIRE(sym_baby->GetCPU().state.task_from_partner[i] == 0);
       }
     }
     sym_baby.Delete();
   }
 
-  SECTION("Symbiont comparison operators")
-  {
+  TEST_CASE("Symbiont comparison operators", "[sgp]") {
```
anyaevostinar, line 65: It looks like this is still nested in the Reproduce test case, I can't see where the { on line 6 closes.

source/test/sgp_mode_test/SGPSymbiont.test.cc
```diff
@@ -56,39 +50,33 @@ TEST_CASE("SGPSymbiont Reproduce", "[sgp]")
...
     emp::Ptr<SGPSymbiont> clone1 = emp::NewPtr<SGPSymbiont>(*sym_parent);
     emp::Ptr<SGPSymbiont> clone2 = emp::NewPtr<SGPSymbiont>(*sym_parent);
     emp::Ptr<SGPSymbiont> different = emp::NewPtr<SGPSymbiont>(&random, &world, &config, CreateNotProgram(99)); // For comparing
 
-    THEN("Symbionts with the same genomes are equal")
-    {
+    THEN("Symbionts with the same genomes are equal") {
```
anyaevostinar, line 70: If you're going to use a THEN, you should make sure to have a WHEN first

anyaevostinar (changes requested):
Looks like the test case nested is causing an error

TahaKhalili:
Run the latest commit, all the errors have been resolved!

source/test/sgp_mode_test/SGPSymbiont.test.cc
```diff
@@ -62,4 +61,68 @@ TEST_CASE("SGPSymbiont Reproduce", "[sgp]") {
...
   }
+}
+
+TEST_CASE("Symbiont comparison operators", "[sgp]") {
+  emp::Random random(31);
+	SymConfigSGP config;
+	SGPWorld world(random, &config, LogicTasks);
+	emp::Ptr<SGPSymbiont> sym_parent = emp::NewPtr<SGPSymbiont>(&random, &world, &config, CreateNotProgram(100));
+
+  WHEN("Parental task tracking is on") {
```
anyaevostinar, line 72: We decided to remove the When/Then since they don't make sense for this test

anyaevostinar (approved):

PR #259, opened 2025-07-30
Complex syms clean

TahaKhalili:


SymSettings.cfg
```diff
@@ -0,0 +1,75 @@
+### MAIN ###
```
anyaevostinar, line 1: You should remove this file, the goal is that all the settings files are in the folder now instead

source/test/sgp_mode_test/ParentTask.test.cc
```diff
@@ -0,0 +1,76 @@
+#include "../../sgp_mode/GenomeLibrary.h"
```
anyaevostinar, line 1: This file should be deleted

anyaevostinar (approved):

PR #283, opened 2025-12-09
Changed to default AVIDA instructions

EliasH-E:
Changed the library of instruction to the same instructions used in default AVIDA plus the minimum to make Symbulation function(Nand, reproduce, sharedio, global anchor)

Quick note, getting print code set up was a little complicated. Some of AVIDA's instructions are a bit longer then Symbulation so I had to up the line length to 17 instead of 12. There probably is a better fix but not one I have found yet.


anyaevostinar (changes requested):
Great, just a couple changes

PR #284, opened 2025-12-12
Restructured Tests, Began filling holes in tests

EliasH-E:
Note: This PR is not done but I wanted it to be visible to allow for easy commenting on current restructuring

-Restructured all sgp_mode tests into the 3 categories of unit, functional and integration.
- unit tests are named after file they originate from
- functional tests are named after the file they are testing or pairs of files that are tested
- integration tests are named after what conceptually is being tested
- Added tests for VT and HT


source/catch/main.cc
```diff
@@ -28,16 +28,30 @@
...
+
+#include "../test/sgp_mode_test/functional_tests/GenomeLibrary_SGPWorld.test.cc"
+#include "../test/sgp_mode_test/functional_tests/HealthHost_Instructions.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPDataNodes.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPHost_Tasks.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPHost.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPSymbiont_Tasks.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPSymbiont.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPWorld_StressHost.test.cc"
+#include "../test/sgp_mode_test/functional_tests/SGPWorld.test.cc"
```
anyaevostinar: Similar to the other comments, if this is a specific complicated thing, it would be good to be more specific in the name of the file

PR #285, opened 2025-12-12
Sym-interaction-value-based evolving death chance for stress mode

K-Johnson-Horrigan:
Symbiont interaction value determines host death chance if tasks match. If IV is positive, death chance is reduced; if IV is negative, death chance is increased.


anyaevostinar (changes requested):
Looks great, just need to move the tests

anyaevostinar (approved):

PR #292, opened 2026-05-12
Nutrient refactor

EliasH-E:
Finished Nutrient Additions:
Split Nutrient interaction into host and sym functors
Changed functors from applying points to calculating them
Created new functor for applying points to hosts.


source/sgp_mode/SGPHost.h
```diff
@@ -43,6 +43,9 @@ class SGPHost : public Host {
   */
   const emp::Ptr<world_t> my_world;
 
+ 
+  size_t current_update = 0;
```
anyaevostinar: Empirical provides GetUpdate, so you can use that directly, sorry for forgetting about that!

source/sgp_mode/SGPHost.h
```diff
@@ -63,6 +66,7 @@ class SGPHost : public Host {
   // }
 
 public:
+  size_t match_count = 0;
```
anyaevostinar: Probably should have this be private and have accessor/mutator for it. Also remember to document it carefully

source/sgp_mode/SGPHost.h
```diff
@@ -252,10 +256,15 @@ class SGPHost : public Host {
...
+
+    match_count = my_world->GetHostSymMatchCount(*this);
+
+    //Gen random number
+    current_update += 1; 
     emp::vector<emp::Ptr<Organism>>& syms = GetSymbionts();
     for (size_t endosym_i = 0; endosym_i < syms.size(); ++endosym_i) {
-      emp_assert(!(syms[endosym_i]->IsHost()));
-      emp::Ptr<sgp_sym_t> cur_symbiont = static_cast<sgp_sym_t*>(syms[endosym_i].Raw());
+      emp_assert(!(syms[(endosym_i + current_update) % syms.size()]->IsHost()));
```
anyaevostinar: This calculation is done twice and confusing to follow so it should definitely be a variable with a comment explaining what is going on (i.e. that this ensures that which sym starts rotates)

source/sgp_mode/SGPHost.h
```diff
@@ -417,42 +426,45 @@ class SGPHost : public Host {
...
 
-          my_world->GetHostTaskSuccesses()[task_id] += 1;
+        double new_points = task_req_info.fun_calc_task_val(
+          task_env,
+          task_req_info,
+          GetPoints()
+        );
+        double task_points = new_points - GetPoints();
+
+        my_world->ApplyHostPoints(*this, task_points,task_id);
```
anyaevostinar, line 463: Let's add a comment here pointing out that the world handles if any points are removed or added by the symbionts

source/sgp_mode/SGPWorld.cc
```diff
@@ -466,9 +466,12 @@ void SGPWorld::ProcessSymOutputBuffer(sgp_sym_t& sym) {
           sym.GetPoints()
         );
         double task_points = new_points - sym.GetPoints();
+        //Hardcode nutrients getting less points
+
+
         // Apply nutrient interaction (if any have been configured) to points
         // NOTE - can inject nutrient interaction here to modify points?
-        task_points = fun_apply_nutrient_interaction(sym, task_points, task_id);
+        //task_points = fun_apply_nutrient_interaction(sym, task_points, task_id);
```
anyaevostinar: You can probably just delete this if we've moved away from it, or maybe make a TODO to delete it once we've settled on the new way. Also probably remove the comment earlier about hardcoding nutrients getting less points

source/sgp_mode/SGPWorld.h
```diff
@@ -529,6 +548,7 @@ class SGPWorld : public SymWorld {
   void SetupReproduction();
   void SetupSymReproduction();
   void SetupHostReproduction();
+  void SetupHostTaskAwards();
```
anyaevostinar: Rewards? What is Awards?

source/sgp_mode/SGPWorld.h
```diff
@@ -626,12 +646,37 @@ class SGPWorld : public SymWorld {
    * Output: The value of the task after applying nutrient interaction.
    * Purpose: To apply the configured nutrient interaction for the given symbiont and task
```
anyaevostinar: Make sure to update documentation to make clear that it isn't applying anything, no side effects

source/sgp_mode/SGPWorld.h
```diff
@@ -626,12 +646,37 @@ class SGPWorld : public SymWorld {
...
+    size_t task_id,
+    size_t symCount
+  ) {
+    return fun_calc_host_nutrient_interaction(host,sym, task_value_before, task_id,symCount);
+  }
+
+   /**
+   * Input: A symbiont, the value of a task before applying nutrient interaction, and the task id.
+   * Output: The value of the task after applying nutrient interaction.
+   * Purpose: To apply the configured nutrient interaction for the given symbiont and task
```
anyaevostinar: ditto

source/sgp_mode/SGPWorld.cc
EliasH-E: I am going to remove the task_points comment however I am going to keep the other one for a brief moment as I completely forgot that I still need to implement that. Currently there is no way for the symbionts to get less points then they normally would like we discussed on Tuesday.

EliasH-E: Okay I have implemented change. Resolving.

source/sgp_mode/SGPWorld.h
EliasH-E: Yes that should be Rewards I believe when typing this I accidentally mixed up the meaning of those two words.

source/sgp_mode/SGPWorld.h
```diff
@@ -626,12 +646,37 @@ class SGPWorld : public SymWorld {
...
+    sgp_sym_t& sym,
+    double task_value_before,
+    size_t task_id,
+    size_t symCount
   ) {
-    return fun_apply_nutrient_interaction(sym, task_value_before, task_id);
+    return fun_calc_sym_nutrient_interaction(host,sym, task_value_before, task_id,symCount);
+  }
+
+  void ApplyHostPoints(
```
anyaevostinar, line 674: Documentation needed

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -14,7 +14,21 @@
 
 // TODO - assert that sym / host has program
 namespace sgpmode {
-
+size_t SGPWorld::GetHostSymMatchCount(sgp_host_t& host){
```
anyaevostinar: Documentation needed

anyaevostinar: I'm conflicted on whether this should be a World method or a Host method. It feels like it's the host's job to check how many of its symbionts match it

anyaevostinar: Even though it is calling a lot of world functors, I think it should be in the host since it is specific to an individual host

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -282,7 +297,12 @@ void SGPWorld::SetupHealthInteractions() {
         emp_assert(donate_prop <= 1.0 && donate_prop >= 0.0);
         const double sym_cycles = (double)sym_state.GetCPUCyclesToExec();
         // How much will sym donate?
+        
         size_t sym_donate = (size_t)(((double)interact) * (donate_prop * sym_cycles));
+        if(host.match_count > 0){
```
anyaevostinar: We need to be careful to make sure that that variable doesn't get out of date. I think the method that updates the match_count should be called "UpdateSymMatchCount" and should be called at the start of each update, then there should be a GetMatchCount which is just an accessor for match_count, but would allow for us to have a check mechanism in the future if we need it

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -320,7 +340,15 @@ void SGPWorld::SetupHealthInteractions() {
         emp_assert(steal_prop <= 1.0 && steal_prop >= 0.0);
         // How much?
         const double host_cycles = (double)host_state.GetCPUCyclesToExec();
-        const size_t sym_steal = (size_t)(((double)interact) * (steal_prop * host_cycles));
+
+        size_t sym_steal_start = (size_t)(((double)interact) * (steal_prop * host_cycles));
+        if(host.match_count > 0){
+          sym_steal_start = (size_t)(((double)interact) * (steal_prop * host_cycles)/host.match_count);
+          if(interact){
+            host.match_count -= 1;
```
anyaevostinar: Since this is a variable that changes throughout the update as symbionts do their interaction, I think we should name it differently so that it is clear what it actually is. Something like "matching_syms_to_interact_with" and then an equivalent (long) accessor name

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -647,60 +678,65 @@ void SGPWorld::SetupNutrientInteractions() {
...
     // Nutrient mutualist - if mutualist task matches a host,
     //  mutualist donates some resources to the host.
-    fun_apply_nutrient_interaction = [this](
+    fun_calc_sym_nutrient_interaction = [this](
+      sgp_host_t& host,
       sgp_sym_t& sym,
       double task_points,
-      size_t task_id
+      size_t task_id,
+      size_t symCount
```
anyaevostinar: switch to snake case and think about if there is a longer name that is more clear on what exact syms we are counting

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -647,60 +678,65 @@ void SGPWorld::SetupNutrientInteractions() {
...
+       double to_donate = std::clamp(
           sgp_config.NUTRIENT_DONATE_PROP() * task_points,
           0.0,
           task_points
         );
+        to_donate = std::min(to_donate, sym.GetPoints());
         // Donate points to host
-        host.AddPoints(to_donate * sgp_config.NUTRIENT_INTERACTION_MULTIPLIER());
+        //host.AddPoints(to_donate * sgp_config.NUTRIENT_INTERACTION_MULTIPLIER());
         // Return task points minus donated value for symbiont to earn
```
anyaevostinar: Update the comment

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -647,60 +678,65 @@ void SGPWorld::SetupNutrientInteractions() {
...
         // Task match, donate proportion of earned task points to host.
         // Can't donate more than task value or less than 0.0
-        const double to_donate = std::clamp(
+       double to_donate = std::clamp(
           sgp_config.NUTRIENT_DONATE_PROP() * task_points,
           0.0,
           task_points
         );
+        to_donate = std::min(to_donate, sym.GetPoints());
         // Donate points to host
```
anyaevostinar: Remove this comment

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -709,58 +745,199 @@ void SGPWorld::SetupNutrientInteractions() {
           0.0,
           task_points
         );
-        // Can't take more than host has
-        const double from_host = emp::Min(
-          host.GetPoints(),
-          to_steal
+        return (to_steal * sgp_config.NUTRIENT_INTERACTION_MULTIPLIER())/symCount;
```
anyaevostinar: Could we switch the signs so that parasites return a negative and mutualists return a positive, it just seems weird this way

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -709,58 +745,199 @@ void SGPWorld::SetupNutrientInteractions() {
...
+     
+
+      const emp::BitVector& sym_task_profile = fun_get_sym_task_profile(sym);
+      bool sym_performed = sym_task_profile.Get(task_id);
+      const double sym_interaction_value = sym.GetIntVal();
+      emp_assert(-1.0 <= sym_interaction_value && sym_interaction_value <= 1.0);
+      if (!sym_performed && (sym_interaction_value > 0.0)) {
+        // Task mismatch from mutualist
+        return 0.0;
+      } else if (!sym_performed && (sym_interaction_value < 0.0)) {
```
anyaevostinar: Could the two !sym_performed be collapsed together?

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -709,58 +745,199 @@ void SGPWorld::SetupNutrientInteractions() {
...
+    exit(-1);
+  }
+
+  
+
+
+
+
+
+  if (GetNutrientSymType() == nutrient_sym_mode_t::MUTUALIST) {
```
anyaevostinar, line 777: Is this duplicate with above? I'm confused

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -709,58 +745,199 @@ void SGPWorld::SetupNutrientInteractions() {
...
+        );
+        return (-to_steal)/symCount;
       }
     };
   } else if (GetNutrientSymType() == nutrient_sym_mode_t::INTERACTION_VALUE_BASED) {
     // Symbiont interaction value determines whether interaction is parasitic (negative)
     //   or mutualistic (positive). Interacition intensity is scaled by symbiont's
     //   interaction value.
-    fun_apply_nutrient_interaction = [this](
+    fun_calc_host_nutrient_interaction = [this](
```
anyaevostinar, line 833: I'm not looking through these, so let me know if they aren't actually duplicated

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -709,58 +745,199 @@ void SGPWorld::SetupNutrientInteractions() {
...
+  }
+
+  
+
+
+
+
+
+  if (GetNutrientSymType() == nutrient_sym_mode_t::MUTUALIST) {
+    // Nutrient mutualist - if mutualist task matches a host,
```
anyaevostinar, line 778: The amount of duplication between host and sym calc functions hurts my soul, we definitely need to figure out what can be streamlined

anyaevostinar: I think most of the comments from the host version apply to the sym version as well

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -769,19 +946,15 @@ void SGPWorld::SetupNutrientInteractions() {
...
+        
         // NOTE - subtract to_steal or from_host?
         // const double from_world = task_points - to_steal;
         // Take points from host
-        return task_points * sgp_config.NUTRIENT_INTERACTION_MULTIPLIER();
+        return (-to_steal)/symCount;
       }
       // Otherwise, no interaction. Return task points.
-      return task_points;
+      return 0;
```
anyaevostinar, line 874: Why 0 instead of task_points?

source/sgp_mode/SGPWorldSetup.cc
EliasH-E: I just wanna check before I make the change, this is in the functor that calculates how many points the parasite gets? So we want it to be if a parasite steals 5 points it should return -5 and then it should run sym.AddPoints(-5 *(-1)) and if it is a mutualist and it donates 5 points it should return 5 and then it should run sym.AddPoints(5 * (-1))

source/sgp_mode/SGPWorldSetup.cc
EliasH-E: I believe that it is 0 points because these functions do not return the total amount of points that the host get, they return how much they get/lose from said symbiont. Thus if there is no interaction it returns 0 as that symbiont has not affected that host in any way.

anyaevostinar: Makes sense, let's update the comment to make that clear

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -922,6 +1097,68 @@ void SGPWorld::SetupSymReproduction() {
   // TODO - anything else to setup here?
 }
 
+void SGPWorld::SetupHostTaskAwards() {
```
anyaevostinar: Documentation

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -922,6 +1097,68 @@ void SGPWorld::SetupSymReproduction() {
...
+      sgp_host_t& host,
+      double task_value_before,
+      size_t task_id
+    ) {
+        int matchCount = 0;
+        emp::vector<emp::Ptr<Organism>>& syms = host.GetSymbionts();
+        for (size_t endosym_i = 0; endosym_i < syms.size(); ++endosym_i) {
+        
+          emp::Ptr<sgp_sym_t> cur_symbiont = static_cast<sgp_sym_t*>(syms[endosym_i].Raw());
+         // bool dead = cur_symbiont->GetDead();
```
anyaevostinar: Why is this commented out? Seems important?

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -922,6 +1097,68 @@ void SGPWorld::SetupSymReproduction() {
...
+          for (size_t endosym_i = 0; endosym_i < syms.size(); ++endosym_i) {
+        
+            emp::Ptr<sgp_sym_t> cur_symbiont = static_cast<sgp_sym_t*>(syms[endosym_i].Raw());
+           // bool dead = cur_symbiont->GetDead();
+            // Skip if dead
+           // if (dead) {
+           //   continue;
+           // }
+
+           const emp::BitVector& endosym_task_profile = fun_get_sym_task_profile(*cur_symbiont);
```
anyaevostinar, line 1075: Seems like a small helper function in the world that just says whether a sym did a task when given the symbiont and task_id would cut down on duplication and improve readability

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -922,6 +1097,68 @@ void SGPWorld::SetupSymReproduction() {
...
+          //  continue;
+         // }
+
+          const emp::BitVector& endosym_task_profile = fun_get_sym_task_profile(*cur_symbiont);
+          bool sym_performed = endosym_task_profile.Get(task_id);
+          matchCount += sym_performed;
+          
+        }
+
+        double totalPoint = 0;
```
anyaevostinar: should be total_points

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -922,6 +1097,68 @@ void SGPWorld::SetupSymReproduction() {
...
+        host.AddPoints(task_value_before);
+    };
+  }
+  else{
+    fun_apply_host_points = [this](
+      sgp_host_t& host,
+      double task_value_before,
+      size_t task_id
+    ) {
+        int matchCount = 0;
```
anyaevostinar: should be matching_sym_count

source/sgp_mode/SGPWorldSetup.cc
anyaevostinar: Or maybe something more descriptive, like "points_from_syms"

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -1093,17 +1331,33 @@ void SGPWorld::SetupHostSymInteractions() {
...
+    size_t symCount
+  ) {
+    return 0.0;
+  };
+
   if (sgp_config.ENABLE_NUTRIENT()) {
     SetupNutrientInteractions();
   }
 
+  SetupHostTaskAwards();
```
anyaevostinar: Definitely should be "Rewards"

source/test/sgp_mode_test/unit_tests/SGPSymbiont.test.cc
```diff
@@ -130,27 +130,31 @@ TEST_CASE("SGPSymbiont Normal Nutrient Parasite without multiplier", "[sgp][sgp-
...
+      host->SetPoints(initial_points);
+      double host_result = world.CalcHostNutrientInteraction(*host, *sym, 8.0, not_task_id,1);
+      double sym_result = world.CalcSymNutrientInteraction(*host, *sym, 8.0, not_task_id,1);
+      
+      
 
-      THEN("Symbiont receives expected amount and host loses the amount") {
-        REQUIRE(result == (sym_score* config.NUTRIENT_INTERACTION_MULTIPLIER()));
-        REQUIRE(host->GetPoints() == initial_host_points - expected_transfer);
+      THEN("Symbiont keep the expected score and host receives the donate amount") {
```
anyaevostinar: "keeps"

source/test/sgp_mode_test/unit_tests/SGPSymbiont.test.cc
```diff
@@ -130,27 +130,31 @@ TEST_CASE("SGPSymbiont Normal Nutrient Parasite without multiplier", "[sgp][sgp-
...
+        REQUIRE(host_result == -expected_transfer);
+        REQUIRE(sym_result == expected_transfer);
       }
     }
   }
 }
 
 //Note, this can't be easily combined with the previous test, because the functor is created at setup, so you can't change the nutrient type on the fly, is that okay for our long term plans?
-TEST_CASE("SGPSymbiont Normal Nutrient mutualist without multiplier", "[sgp][sgp-functional]") {
+TEST_CASE("SGPSymbiont Normal Nutrient mutualist without multiplier", "[sgp-1][sgp-functional]") {
```
anyaevostinar: remove the -1

anyaevostinar (changes requested):

source/sgp_mode/SGPWorldSetup.cc
anyaevostinar: No, you're right, if it's calculating what the symbiont gets, it makes sense to be positive for parasites and negative for mutualists, probably a comment would be helpful for future me that is confused again.

source/sgp_mode/SGPHost.h
```diff
@@ -585,6 +592,22 @@ class SGPHost : public Host {
     hardware.Reset(); // NOTE - this function was previously just Initializing state,
                       // which didn't reset the cpu. I think we want to reset the CPU here also?
   }
+
```
anyaevostinar, line 596: Documentation needed

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -1114,25 +1029,25 @@ void SGPWorld::SetupHostTaskAwards() {
       double task_value_before,
       size_t task_id
     ) {
-        int matchCount = 0;
+        int task_matching_sym_count = 0;
         emp::vector<emp::Ptr<Organism>>& syms = host.GetSymbionts();
         for (size_t endosym_i = 0; endosym_i < syms.size(); ++endosym_i) {
```
anyaevostinar, line 1049: Why isn't this using the Host's UpdateMatching syms method?

source/sgp_mode/SGPWorldSetup.cc
```diff
@@ -1145,15 +1060,15 @@ void SGPWorld::SetupHostTaskAwards() {
            const emp::BitVector& endosym_task_profile = fun_get_sym_task_profile(*cur_symbiont);
            bool sym_performed = endosym_task_profile.Get(task_id);
            if(sym_performed){
-              size_t sym_task_point = CalcSymNutrientInteraction(host,*cur_symbiont, task_value_before, task_id,matchCount);
-              totalPoint += CalcHostNutrientInteraction(host, *cur_symbiont, task_value_before, task_id,matchCount);
+              size_t sym_task_point = CalcSymNutrientInteraction(host,*cur_symbiont, task_value_before, task_id,task_matching_sym_count);
+              point_differnce_from_syms += CalcHostNutrientInteraction(host, *cur_symbiont, task_value_before, task_id,task_matching_sym_count);
```
anyaevostinar: typo, "point_difference_from_syms"

source/test/sgp_mode_test/functional_tests/NutrientMode.test.cc
```diff
@@ -0,0 +1,113 @@
...
+    WHEN("Parasite steals from host") {
+      
+      sym->SetPoints(initial_points);
+      host->SetPoints(initial_points);
+      double host_result = world.CalcHostNutrientInteraction(*host, *sym, 8.0, not_task_id,1);
+      double sym_result = world.CalcSymNutrientInteraction(*host, *sym, 8.0, not_task_id,1);
+      
+      
+
+      THEN("Symbiont keep the expected score and host receives the donate amount") {
```
anyaevostinar: 'keeps'

source/test/sgp_mode_test/functional_tests/NutrientMode.test.cc
```diff
@@ -0,0 +1,113 @@
...
+        REQUIRE(host_result == -expected_transfer);
+        REQUIRE(sym_result == expected_transfer);
+      }
+    }
+  }
+}
+
+//Note, this can't be easily combined with the previous test, because the functor is created at setup, so you can't change the nutrient type on the fly, is that okay for our long term plans?
```
anyaevostinar: It should be possible to combine by putting the share configs before the GIVEN and then having whatever is different inside a GIVEN and worldsetup called within the GIVEN, since everything is rerun, so it will be a new world and random, even if they are shared code

anyaevostinar (changes requested):
A few more things to fix

source/sgp_mode/SGPWorldSetup.cc
EliasH-E: The main difference between the two is that this section of code checks how many symbionts the host has that have completed the specific task the host just did, while the UpdateMatching syms method calculates the total number of syms that match at all with the host. If we wanted to, we could also change the nutrient mode to function like that but it is fundamentally different if we allow multiple task credit.

anyaevostinar (approved):

PR #293, opened 2026-05-18
2026-05-18 main patches

amlalejini:
A few minor patches:
- Adjust `before_host_process_sig` signal trigger timing to occur as it did previously
- Adjust initial not program in program builder to fully cycle through all inputs in the input buffer when looped over
- Add a few local directories to .gitignore
- Set verbose arg to true in Run function call for sgp mode

### DIFF
--- a/source/sgp_mode/SGPHost.h
+++ b/source/sgp_mode/SGPHost.h
@@ -271,8 +271,7 @@ public:
         *this
       );
     }
-    //TODO AEV: rename before_host_process_sig to before_host_exec_cpu_cycles_sig since that's what it really is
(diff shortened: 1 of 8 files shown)


source/sgp_mode/SGPHost.h
```diff
@@ -282,7 +272,7 @@ class SGPHost : public Host {
       );
     }
 
-    
+    my_world->before_host_process_sig.Trigger(*this);
```
anyaevostinar: Rename signal to "before_host_cpu_execute_sig"?

anyaevostinar (approved):

PR #295, opened 2026-05-21
Hardcoded temporally changing environments

K-Johnson-Horrigan:
- according to some interval, tasks switch between being rewarded (+ points) or punished (- points). Task groups: NOT, AND, OR start rewarded. NAND, AND-NOT, OR-NOT start punished.
- SenseTask instruction determines whether a value in a register will be rewarded or punished if processed.


source/test/sgp_mode_test/unit_tests/Instructions.test.cc
```diff
@@ -780,3 +775,230 @@ TEST_CASE("Test freeliving symbiont instructions", "[sgp]") {
   }
 
 }
+
+
+TEST_CASE("Test SenseTask instruction", "[sgp]"){
```
anyaevostinar: This is a very long test and really doesn't seem like a unit test given how complex it is. That is partially because we really want helper functions to allow for smaller unit tests of specific behavior ultimately. For now, I think that this megatest should just be its own file in the functional_tests since it is testing the interaction between tasks and SenseTask. That will then allow you to break it up into several small tests for each of the scenarios and have small "when"/"then" sections that are closer to describing what is actually happening

source/test/sgp_mode_test/unit_tests/Instructions.test.cc
```diff
@@ -780,3 +775,230 @@ TEST_CASE("Test freeliving symbiont instructions", "[sgp]") {
...
+      sym_hw.SetRegisters({7, 12, 9}); // Initial register values
+
+      THEN("SenseTask puts a 0 into register 1"){
+        // Run symbiont program
+        sym_hw.RunCPUStep(1); // Anchor
+        REQUIRE(sym_hw.GetRegister(0) == 7);
+        REQUIRE(sym_hw.GetRegister(1) == 12);
+        REQUIRE(sym_hw.GetRegister(2) == 9);
+
+        sym_hw.RunCPUStep(1); //  io   r0 // puts input1 into a
```
anyaevostinar: I get the desire to test each step, but it means that many of these requires are not actually testing SenseTask, they are testing all the many other instructions being used. Cut down to just testing SenseTask in this test. The other functionally should be tested in its own tests, so that if this one fails, we know it's specifically because of SenseTask

source/test/sgp_mode_test/unit_tests/TempChangingEnvironments.test.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "../../../sgp_mode/SGPWorldSetup.cc"
+#include "../../../sgp_mode/SGPWorldData.cc"
+#include "../../../sgp_mode/ProgramBuilder.h"
+
+
+
+#include "../../../catch/catch.hpp"
+
+
+TEST_CASE("Temporally changing environment test", "[sgp]") {
```
anyaevostinar: This is definitely not a unit test, it should at minimum be moved to functional, but we could also consider if it's time to make a higher level. I think functional is fine though. Also you should break this up into multiple tests instead of one mega test. Each individual scenario should be its own test. If you hate the duplicate set up code, you can make a little helper function for that part.

anyaevostinar (changes requested):
The tests should be moved to functional_tests and broken up and streamlined.

anyaevostinar (approved):
Awesome!

PR #298, opened 2026-05-29
Updated stress mode test

LaurenC267:


source/test/sgp_mode_test/functional_tests/SGPHardware.test.cc
```diff
@@ -0,0 +1,65 @@
+#include "emp/math/Random.hpp"
```
anyaevostinar: We don't actually need this file anymore, delete

anyaevostinar (approved):

PR #304, opened 2026-06-06
SGP test updates (cont x3) into dev-refactor-again

K-Johnson-Horrigan:
old_sgp_mode_test/functional_tests/SGPDataNodes_SGPWorld.test.cc
- renamed to SGPWorldData_SGPWorld.test + updated to reflect current data files/tracking
- added unit test to check that CreateDataFiles creates data files
- renamed to SGPWorldDaata_SGPWorld

old_sgp_mode_test/unit_tests/HealthHost.test.cc
- Before_endosym_host_process_sig appears to be called twice, once in host Process() and once in symbiont Process(). Deleted the symbiont Process call; keeping it in host lets us do steal/donation of cycles before host starts executing cycles.
- Renamed to functional/HealthMode 

need_updating/HealthHost.test.cc
- Changed the numbers, but captured the behavior that evolving worlds do more NOTs than non-evolving worlds

Other notes
- commented out free living syms test in functional_tests/SGPWorld.test.cc + restored the assertion in sgpworlddata that fails when FLS is on


source/sgp_mode/SGPSymbiont.h
```diff
@@ -214,7 +214,6 @@ class SGPSymbiont : public Symbiont {
     }
 
     GetHardware().GetCPUState().SetLocation(pos);
-    if(my_host) my_world->TriggerBeforeEndoSymProcessSig(pos, *this, my_host);
```
anyaevostinar: I realized that there are actually two different signals and I had mixed them up when doing this porting, so this shouldn't be deleted since the call within host should be to before_endosym_host_process whereas this one is to before_endosym_process. I'm pushing a fix currently

anyaevostinar: I've fixed it in your branch

anyaevostinar (approved):

anyaevostinar (approved):

PR #309, opened 2026-07-06
Tag-based SGP interaction

K-Johnson-Horrigan:
This PR renames TASK_PROFILE_COMPATIBILITY_MODE to INTERACTION_PROFILE_COMPATIBILITY_MODE and implements a wrapper function that allows tags to be used for interaction. However, tags cannot be used for horizontal transmission, and thus the horizontal transmission settings automatically pass when tag matching is on in SGP. Maybe worth keeping both the old TASK_PROFILE_COMPATIBILITY_MODE option and the new INTERACTION_PROFILE_COMPATIBILITY_MODE  option, for backcompatibility/task based horizontal transmission; however, that would contribute to config bloat.


source/sgp_mode/SGPConfigSetup.h
```diff
@@ -26,8 +26,8 @@ EMP_EXTEND_CONFIG(SymConfigSGP, SymConfigBase,
...
-  VALUE(TASK_PROFILE_COMPATIBILITY_MODE, std::string, "always", "How is compatibility determined for task profiles? always, task-any-match, task-perfect-match"),
-  VALUE(HORIZONTAL_TRANSMISSION_COMPATIBILITY_MODE, std::string, "always", "How is compatibility determined for horizontal transmission? Relies on TASK_PROFILE_COMPATIBILITY_MODE being set appropriately. Options are: always, task-profile-compatible, task-profile-strictly-stronger-match, task-profile-stronger-or-equal-match"),
+  VALUE(INTERACTION_PROFILE_COMPATIBILITY_MODE, std::string, "always", "How is compatibility determined for task profiles? always, task-any-match, task-perfect-match, tag-probabilistic-match"),
+  VALUE(HORIZONTAL_TRANSMISSION_COMPATIBILITY_MODE, std::string, "always", "How is compatibility determined for horizontal transmission? Relies on INTERACTION_PROFILE_COMPATIBILITY_MODE being set appropriately. Options are: always, task-profile-compatible, task-profile-strictly-stronger-match, task-profile-stronger-or-equal-match"),
```
anyaevostinar, line 30: Should the options change "interaction-profile-compatible" etc?

source/sgp_mode/SGPW_TaskProfileSetup.cc
```diff
@@ -48,34 +48,81 @@ void SGPWorld::SetupTaskProfileMode() {
   }
 }
 
+void SGPWorld::SetupInteractionCompatibilityMode() {
```
anyaevostinar, line 51: More complete docstring needed

source/sgp_mode/SGPW_TaskProfileSetup.cc
```diff
@@ -48,34 +48,81 @@ void SGPWorld::SetupTaskProfileMode() {
...
-  } else if (sgp_config.TASK_PROFILE_COMPATIBILITY_MODE() == "task-perfect-match") {
+  } else if (sgp_config.INTERACTION_PROFILE_COMPATIBILITY_MODE() == "task-perfect-match") {
     fun_task_profile_compatibility_check = [this](
       const emp::BitVector& a,
       const emp::BitVector& b
     ) -> bool {
       return a == b;
     };
+  } else if(sgp_config.INTERACTION_PROFILE_COMPATIBILITY_MODE() == "tag-probabilistic-match"){
+    fun_task_profile_compatibility_check = [this](
```
anyaevostinar, line 118: This feels risky since it seems like we might use task_profile compatibility in some way even when the interaction is based on tags, which I guess would mean we want a separate config for task compatibility still

source/sgp_mode/SGPConfigSetup.h
```diff
@@ -26,8 +26,8 @@ EMP_EXTEND_CONFIG(SymConfigSGP, SymConfigBase,
   VALUE(INTERACTION_MECHANISM, std::string, "default", "What sgp organisms should population the world? (Options: 'default')"),
   VALUE(VT_TASK_MATCH, bool, false, "Should task matching be required for vertical transmission? (0 for no, 1 for yes)"),
   VALUE(TASK_PROFILE_MODE, std::string, "self-all", "What should we use for task profiles for host-symbiont compatibility, preferential ousting, etc.? Options are parent-all, self-all, self-first, parent-first. 'all' means all tasks, 'first' means only first task performed."),
```
anyaevostinar, line 28: How does task_profile mode interact with interaction profile compatbility when it's tag based? I think we need to think about this system overall a bit more to try to make this intuitive, oof

PR #310, opened 2026-07-07
Implement JSON features for ProgramBuilder

mmore500:


source/sgp_mode/hardware/Instructions.h
```diff
@@ -44,6 +46,14 @@ namespace sgpmode::inst {
     }                                                                          \
     static size_t prevalence() { return 1; }                                   \
     static std::string name() { return #InstName; }                            \
+    template<typename Spec>                                                    \
```
anyaevostinar, line 53: I realize it's unfair since the existing code doesn't have much documentation, but this macro is getting rather complicated, so some documentation on these would be really nice

source/sgp_mode/ProgramBuilder.h
```diff
@@ -32,6 +42,15 @@ class ProgramBuilder {
 
   rectifier_t& rectifier;
 
+  void rectify_with_warning(program_t& program) const {
```
anyaevostinar: Should be Pascal case please, and a doc string would be helpful to clarify what rectify actually means so that we don't have to go into sgp-lite documentation for that explanation

source/sgp_mode/ProgramBuilder.h
```diff
@@ -372,10 +391,47 @@ class ProgramBuilder {
     program.resize(length - 1);
     AddInst(program, repro_op);
     // Remove any deleted instructions
-    program.Rectify(rectifier);
+    rectify_with_warning(program);
     return program;
   }
 
+  program_t ParseJsonString(const std::string& json_str) {
```
anyaevostinar, line 628: Again I realize it's quite unfair since there is very little documentation in this file, but I'd love to start fixing that by adding our typical doc strings now :)

anyaevostinar (changes requested):
Thanks so much for doing this! If possible, it'd be nice to get some of the doc strings in before pulling it in, but just let me know if you don't have time for them right now!

source/sgp_mode/ProgramBuilder.h
mmore500: 31220d0ac90e2dd8b30e1cb7e255864794674c13

source/sgp_mode/hardware/Instructions.h
mmore500: No such thing! 9b63f594d555090a0699c7111ad40992b2388b20

source/sgp_mode/ProgramBuilder.h
mmore500: Tried to match the convention elsewhere in the project. 9b63f594d555090a0699c7111ad40992b2388b20

mmore500:
Review comments look good and should be addressed! Also took advantage of the opportunity to add matching docstrings in touched areas. Some hand-written, and some generated with hand review/revision. Noted in commit messages.

anyaevostinar (approved):
Thanks!!

PR #311, opened 2026-07-07
Fixed jumps and anchors in Genome Printing

EliasH-E:
Fixed error when genomes were printed that caused jump instructions to not have any information on where they jump was to
Fixed error when genomes were printed that caused anchors to be printed with large binary numbers


source/sgp_mode/hardware/SGPHardware.h
```diff
@@ -303,16 +303,19 @@ void SGPHardware<HW_SPEC_T>::PrintOp(
...
       out << "    " << emp::to_lower(name);
       for (size_t i = 0; i < 12 - name.length(); i++) {
         out << ' ';
       }
       out << 'r' << (int)ins.args[0] << ", r" << (int)ins.args[1] << ", "
           << tag_name;
+
     } else if (name == "Global Anchor") {
-      out << tag_name << " " << ins.tag << ':';
+      //out << tag_name << " " << ins.tag << ':';
```
anyaevostinar: Remove dead code

anyaevostinar (changes requested):
Great job! One tiny removal and it's good to go

anyaevostinar (approved):

PR #312, opened 2026-07-08
kelleyde sgp-copy: Event System, TaskValue Event Tests, CreateNANDProgram

kelleyde:
Event System
Event files path: source/sgp_mode/events

Event.h - generic event class that more specific event type classes can inherit from. The generic event class includes the minimum necessary member variables for creating an event. Member functions have logic for determining if an event is one time or reoccurring, and for advancing reoccurring events to the next step. 

ExampleEvent.h - example of an event class. It is not meant to be functional, just a guide to creating new event classes and types. 

TaskValueEvent.h - task_value event class and data structures. Task value events are events where the value(s) of a certain logic task(s) are altered/changed. They have three different action types (specified in the “action” field of the json file), CHANGE will set a task value to a new value, ADD will add a specific value to the previous task value, or MULT will multiply the previous task value by a specific value. These changes can be applied to symbiont tasks (SYM), host tasks (HOST), or both (SHARED), depending on the “group” field of the json file. 

Example of json file fields: "event_type": "task_value", "task_name": "NAND",  "action": "change", "value": 50,  "timing": "1",  "group": "host"
Task name can also be a list of multiple different tasks, for example [“NOT”, “NAND”].
If timing is three integers separated by a “:” then it is a reoccurring event, for example “1:4:2” (order is “start:stop:step”). A single integer in the “timing” field indicates a one time event.
Options for “group” are “host”, “symbiont”, “shared”.
Options for “action” are “mult”, “add”, “change”.

EventManager.h - stores all one-time and recurring events in two different vectors, loads events from json file, processes and reorder events, deletes finished events from end of the vector, includes logic for adding new events. For managing all loaded events and processing them.

EventTiming.h - tracks the timing of events, for example, by getting start, end, and next updates, tracking the next step for reoccurring events, or resetting timing. For helping track timing of events.

EventTypeDefinition.h - contains necessary variables and functions to define an event type and includes handling for loading and processing events. For defining event types.

EventTypeLibrary.h - contains map of event names to event ids and vector of event type definitions. There are functions for checking valid event types, loading and processing different event types, adding new event types, getting event ids. For storing event types.

In SGPWorld.h, Update( ) line 656, events are processed at the start of every update.
In SGPWorldSetup.cc line 50 and 1208, events are loaded in from a json file. 

Task Value Event Tests
Path: source/test/sgp_mode_test/functional_tests/TaskValueEvent.test.cc
Tests different task value event actions (change, add, mult), groups (host, symbiont, shared), and timing (one time, reoccurring). Also tests task value events on different tasks and multiple tasks at a time. Tests check change in task value and in points host and symbiont accumulate.

Changes to ProgramBuilder.h CreateNANDProgram line 358. 
Two registers are used to cycle through a NAND b, b NAND c, c NAND d, and d NAND a. New input gets placed in the register that holds the NAND of the previous two inputs. The NAND of the next two inputs are placed in the opposite register of the new input. For example register 0 = a, register 1 = b, and register 0 = a NAND b, then the next input c is put in register 0, and c NAND b goes in register 1. Following this logic, the next new input d goes in register 1, and c NAND d goes in register 0.


source/sgp_mode/events/event_types/Event.h
```diff
@@ -0,0 +1,102 @@
...
+// See ExampleEvent.h for a minimal example of how to implement a new event type.
+class Event {
+protected:
+  size_t event_type_id = 0;       // Event type ID for this event. Filled in when event is loaded from file.
+  std::string event_type{"NULL"}; // Human-readable event type.
+  EventTiming timing;             // Manages event's timing (one-time vs. recurring start/stop/step)
+  bool is_done = false;           // Can be used by an event to end a recurring event before its end update
+
+public:
+  size_t GetEventTypeID() const { return event_type_id; }
```
anyaevostinar, line 108: Need doc strings throughout

source/sgp_mode/events/event_types/Event.h
```diff
@@ -0,0 +1,102 @@
...
+  // Advance next update, return new next update.
+  size_t AdvanceNextUpdate() {
+    timing.Step();
+    return timing.GetNextUpdate();
+  }
+
+};
+
+// Common parsing function used by many event types.
+void SetEventTimingFromJSON(
```
anyaevostinar, line 247: Full doc string definitely needed

source/sgp_mode/events/event_types/Event.h
```diff
@@ -0,0 +1,102 @@
...
+  nlohmann::json& event_json,
+  emp::Ptr<Event> event_ptr
+) {
+  emp_assert(event_json.contains("timing"));
+  const std::string timing_str(event_json["timing"]);
+  // EventTiming timing;
+  // Given as a single (integer) number?
+  if (emp::is_digits(timing_str)) {
+    const size_t start_u = emp::from_string<size_t>(timing_str);
```
anyaevostinar: Fuller variable names will help readablility a lot, such as start_update

source/sgp_mode/events/event_types/ExampleEvent.h
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * Output: Pointer to new event instance created and configured by this load
+   *         function.
+   */
+  template <typename WORLD_T>
+  static emp::Ptr<ExampleEvent> LoadEventFromJSON(
+    json_t& event_json,
+    WORLD_T& world
+  ) {
+    // This should be a task_value event
```
anyaevostinar, line 35: example event?

source/sgp_mode/events/EventManager.h
```diff
@@ -0,0 +1,437 @@
...
+   *
+   * Input: json oject to load an event from, reference to the world
+   *
+   * Output: Pointer to a new event object
+   */
+  emp::Ptr<event_t> LoadEventFromJSON(json_t& event_json, world_t& world) {
+    // Check that event_json has event type
+    emp_assert(event_json.contains("event_type"));
+    const std::string event_type(event_json["event_type"]);
+    // emp::Ptr<Event> loaded_event;
```
anyaevostinar, line 70: dead code?

source/sgp_mode/events/EventManager.h
```diff
@@ -0,0 +1,437 @@
...
+public:
+  /**
+   * Purpose: EventManager destructor. Responsible for cleaning up any event objects.
+   */
+  ~EventManager() {
+    ClearEvents();
+  }
+
+  /**
+   * Purpose:
```
anyaevostinar, line 212: Missing

source/sgp_mode/events/EventManager.h
```diff
@@ -0,0 +1,437 @@
...
+   *          CAUTION: the event manager makes assumptions about the order of one-time
+   *          events based on their timing. This accessor is used primarily for testing.
+   *          If one-time events are modified via this function, you must resort them (ReorderOneTimeEvents)
+   */
+  emp::vector<emp::Ptr<event_t>>& GetOneTimeEvents() {
+    return one_time_events;
+  }
+
+  /**
+   * Purpose: load in and process events configuratoin json file (includes
```
anyaevostinar, line 260: typo

source/sgp_mode/events/EventManager.h
```diff
@@ -0,0 +1,437 @@
...
+   * Purpose: Process any events that need to be processed on the current world
+   *          update.
+   *
+   * Input: Reference to the world.
+   *
+   * Output: None.
+   */
+  void ProcessEvents(world_t& world) {
+    // Get current update in the world. Process all events that should occur on this update.
+    // Options:
```
anyaevostinar, line 307: It looks like this is decided and I agree so this should be removed

source/sgp_mode/events/EventTypeDefinition.h
```diff
@@ -0,0 +1,130 @@
...
+template<typename WORLD_T>
+class EventTypeDefinition {
+public:
+  using event_t = Event;          // event class alias
+  using world_t = WORLD_T;        // world type alias
+  using json_t = nlohmann::json;  // json file type
+
+  // Event handler function type
+  using fun_event_handler_t = std::function<void(
+    world_t&,         // sgp world to use helper functions and avoid circualr dependency
```
anyaevostinar, line 32: typo

source/sgp_mode/tasks/LogicTaskEnvironment.h
```diff
@@ -65,21 +66,11 @@ class LogicTaskEnvironment {
   // TODO - track task performance?
 
   // TODO - move this into util file in json directory
```
anyaevostinar, line 68: This comment is done now right?

source/sgp_mode/SGPConfigSetup.h
```diff
@@ -68,10 +68,11 @@ EMP_EXTEND_CONFIG(SymConfigSGP, SymConfigBase,
...
 
-  GROUP(TEMP_CHANGING_ENVIRONMENT, "Temporally changing environment settings (task rewards change over time)"),
-  VALUE(ENABLE_TEMP_CHANGING_ENVIRONMENT, bool, false, "Do task reward values change over time?"),
-  VALUE(TEMP_CHANGING_ENVIRONMENT_INTERVAL, size_t, 100, "How many updates elapse between task reward value shuffling?"),
-  VALUE(TEMP_CHANGING_ENVIRONMENT_ORG_TYPE, std::string, "static", "Can organisms sense task reward values? (plastic-both: both symbionts and hosts can sense whether tasks are rewarded; static: neither hosts nor symbiont can sense whether tasks are rewarded)"),
+  GROUP(EVENTS, "Events settings"),
+  VALUE(EVENTS_CFG_PATH, std::string, "events.json", "JSON file that provides event configuration"),
+
+  GROUP(INSTRUCTIONS, "Instruction settings"),
+  VALUE(SENSE_TASK_INSTRUCTION, bool, false, "Should sense task instruction be included in the instruction set?"),
```
anyaevostinar, line 75: Probably clearly call it 'SenseTask' so that it's obvious what the name of the instruction is

source/sgp_mode/SGPWorld.h
```diff
@@ -824,6 +837,10 @@ class SGPWorld : public SymWorld {
    */
   void Update() override {
     emp_assert(setup);
+    // NOTE - When do we want events to occur? Typically, I think we want them
```
anyaevostinar, line 840: Agreed, I think they have to just be the first thing that happens in an update. So let's remove the comment

source/test/sgp_mode_test/unit_tests/SGPHost.test.cc
```diff
@@ -257,6 +262,7 @@ TEST_CASE("ProcessOutputBuffer", "[sgp][sgp-unit]") {
   GIVEN("A host with valid values in its input and output buffers") {
     emp::Random random(31);
     sgpmode::SymConfigSGP config;
+    config.EVENTS_CFG_PATH("source/test/sgp_mode_test/no-events.json");
```
anyaevostinar, line 265: It seems like there should be some way to indicate no events file and that should be fine, since it is reasonable to not specify any events and it's annoying to have to have an empty events file to run anything. Is it too gross to have the loading code not error if the "path" specified is "no-events" (without the .json) and just output that no events file was loaded as specified?

source/test/sgp_mode_test/functional_tests/TaskValueEvent.test.cc
```diff
@@ -0,0 +1,437 @@
...
+
+      // create host with NAND operation
+      emp::Ptr<sgp_host_t> host = emp::NewPtr<sgp_host_t>(&random, &world, &config, builder.CreateNandProgram(50));
+      // create symbiont with NAND operation
+      emp::Ptr<sgp_sym_t> symbiont = emp::NewPtr<sgp_sym_t>(&random, &world, &config, builder.CreateNandProgram(50));
+      // add host and sym
+      world.AddOrgAt(host, 0);
+      host->AddSymbiont(symbiont);
+
+      THEN("Events are processed") {
```
anyaevostinar, line 73: Ideally this when/then would have more information about what the test is actually checking, but we can leave that for further dev work in the future

anyaevostinar (changes requested):
Wow, very nice! Couple of places with commented out code or discussion comments that can be cleaned up. Main thing is making it so that you don't have to have an events file to run if you don't want any events

source/sgp_mode/events/EventManager.h
amlalejini: yup, removing!

source/sgp_mode/events/EventManager.h
amlalejini: Yup, removing!

source/sgp_mode/tasks/LogicTaskEnvironment.h
amlalejini: Yup, I think both of those todos are done. Removing those comments.

source/sgp_mode/SGPConfigSetup.h
amlalejini: Good call, totally agree. Just in case there are future enable/disable config options, what about `INCLUDE_INSTRUCTION_SenseTask`? to make it super clear what we're toggling

source/test/sgp_mode_test/unit_tests/SGPHost.test.cc
amlalejini: As per discussion in slack, if no file is found + the config is set to the default setting, we'll generate a default (empty) events file and print a helpful message.

PR #321, opened 2026-07-16
ProcessOutputBuffer Refactor

Gabe-Rose:
I went ahead and did the functor refactoring, and I think I did it all right. I will check data congruency in nutrient-parasite with the main branch asap. Some other refactor thoughts are:

SetupHostTaskRewards(); SetupSymTaskRewards();  -> SetupNutrientInteractions(); (InteractionMechanismSettup)
What would happen if Defaults/all modes in cfg false?

Remove my_world->GetConfigPtr() from SGPWorld.h? There are two options here, and we easily choose one.


source/sgp_mode/SGPW_InteractionMechanismSetup.cc
```diff
@@ -27,10 +30,9 @@ namespace sgpmode {
     // Configure nutrient interactions
     if (sgp_config.ENABLE_NUTRIENT()) {
       SetupNutrientInteractions();
-    } // default is no nutrient interactions, which is set in SGPWorld constructor
-
-    SetupHostTaskRewards();
-
+      OverrideHostRewardsNutrient();
```
anyaevostinar: It seems that these Override calls should just both be in SetupNutrientInteractions since they are the core of nutrient interactions

source/sgp_mode/SGPWorld.h
```diff
@@ -102,6 +102,12 @@ class SGPWorld : public SymWorld {
     size_t     /* task id */
   )>;
 
+  using func_apply_sym_points_t = std::function<void(
```
anyaevostinar: should be "fun", not "func"

source/test/sgp_mode_test/unit_tests/SGPSymbiont.test.cc
```diff
@@ -110,25 +110,41 @@ TEST_CASE("Symbiont Process", "[sgp][sgp-unit]") {
       }
     }
   }
-
 }
 
-// TEST_CASE("SGPSymbiont destructor cleans up shared pointers and in-progress reproduction", "[sgp][sgp-unit]") {
```
anyaevostinar, line 116: Leave this commented test in, we need to fix it eventually

anyaevostinar (changes requested):
A few more things to change

anyaevostinar (approved):

PR #323, opened 2026-07-23
Web GUI PR

EliasH-E:
Fixed default mode web gui
Added functionality for sgp-web gui
Started adding functionality for SFX in web gui


web/SGPSymAnimate.h
anyaevostinar, line 1: I think this file should live within the top-level web folder, source is for only non-web code.

source/web/SGPsymbulation-web.cc
anyaevostinar, line 1: This file should also go in the top level web folder, where the existing symbulation-web.cc is

Makefile
```diff
@@ -38,6 +41,7 @@ default: default-mode
 
 native: default-mode
 web: symbulation.js
+sgp-web: sgp-symbulation.js
```
anyaevostinar, line 44: I feel like we usually do symbulation-sgp most places?

anyaevostinar (changes requested):
Just a few things to move around

source/web/SGPsymbulation-web.cc
EliasH-E: Hi Anya, It currently is in the same folder as symbulation-web.cc. There currently are 3(kinda 4) different web folders. There is a top level one outside of source that holds all of the UI based assets and graphics as well as the js compilation of Symbulation. Then there is the web folder in source, that holds SGPsymbulation-web.cc and symbulation-web.cc. Then there is the web folder in sgp_mode that has SGPSymAnimate which I am happy to move to one of the other web folders, I just do not know which one. Finally there is SymAnimate which just resides in source which I will move to the same location as SGPSymAnimate.

source/web/symbulation-web.cc
anyaevostinar, line 1: It looks like this file got deleted instead of moved, I don't see it in web at all

anyaevostinar (changes requested):
Need to put symbulation-web.cc into web folder by the looks of it.

PR #327, opened 2026-08-06
Creating Accessors, protecting variables and changing CPU state location access

EliasH-E:


source/sgp_mode/hardware/CPUState.h
```diff
@@ -86,7 +86,7 @@ class CPUState {
   emp::Ptr<Organism> organism; // Unowned pointer to organism using this CPU.
   emp::Ptr<world_t> world_ptr; // Unowned pointer to the world using this CPU.
 
-  emp::WorldPosition location;
+  //emp::WorldPosition location;
```
anyaevostinar: Let's just delete these commented out lines right away so that they don't become junk lingering around forever

source/sgp_mode/SGPHost.h
```diff
@@ -81,8 +81,8 @@ class SGPHost : public Host {
   ) :
     Host(_random, _world, _config, _intval, _syms, _repro_syms, _points),
     hardware(_world, this),
-    my_world(_world)
-    // sgp_config(_config)
+    my_world(_world),
+      sgp_config(_config)
```
anyaevostinar: Can you line up the spacing here while you are editing this file?

source/sgp_mode/SGPHost.h
```diff
@@ -171,10 +171,10 @@ class SGPHost : public Host {
    */
   void SetReproCount(size_t _in) { reproductions = _in; }
 
-  void SetLocation(emp::WorldPosition pos) {
-    hardware.GetCPUState().SetLocation(pos);
-    Host::SetLocation(pos);
-  }
+  //void SetLocation(emp::WorldPosition pos) {
```
anyaevostinar: Just delete

source/sgp_mode/SGPWorld.h
```diff
@@ -584,6 +446,150 @@ class SGPWorld : public SymWorld {
       (static_cast<sgp_sym_t*>(org_ptr.Raw()))->GetHardware().GetCPUState();
   }
 
+  // -- Signals to allow custom behavior to be triggered by other classes -- /
+  // AEV question: getter methods for these?
```
anyaevostinar: You can remove my comment now :)

source/test/sgp_mode_test/unit_tests/SGPHost.test.cc
```diff
@@ -309,4 +309,30 @@ TEST_CASE("Check that hosts and syms can't have negative points", "[sgp][sgp-uni
       }
     }
   }
+}
+
+TEST_CASE("CPUState and SGPHost always have the same location","[sgp][elias]"){
```
anyaevostinar: Remove your name tag :)

source/test/sgp_mode_test/unit_tests/SGPHost.test.cc
```diff
@@ -309,4 +309,30 @@ TEST_CASE("Check that hosts and syms can't have negative points", "[sgp][sgp-uni
...
+  using sgp_sym_t = sgpmode::SGPSymbiont<hw_spec_t>;
+
+  emp::Random random(31);
+  sgpmode::SymConfigSGP config;
+  world_t world(random, &config);
+  auto& prog_builder = world.GetProgramBuilder();
+  emp::Ptr<sgp_host_t> host = emp::NewPtr<sgp_host_t>(&random, &world, &config, prog_builder.CreateReproProgram(100));
+  
+  WHEN("Initialized"){
+   REQUIRE(host->GetHardware().GetCPUState().GetLocation().GetIndex() == host->GetLocation().GetIndex());
```
anyaevostinar: It seems weird to have WHEN without THEN. You can have a require outside of a WHEN/THEN, or just put a THEN around the requires so that it reads more clearly

source/test/sgp_mode_test/unit_tests/SGPSymbiont.test.cc
```diff
@@ -132,3 +132,30 @@ TEST_CASE("Symbiont Process", "[sgp][sgp-unit]") {
 //     }
 //   }
 // }
+
```
anyaevostinar, line 135: Same changes here as host tests

source/Organism.h
```diff
@@ -288,14 +288,6 @@ class Organism {
     std::cout << "ClearReproSyms called from Organism" << std::endl;
     throw "Organism method called!";
   }
-  virtual void CycleTransfer(int _in) {
```
anyaevostinar, line 291: Do these not need to be in here anymore? Why were they needed before?

anyaevostinar (changes requested):
A few things to clean up

source/Organism.h
EliasH-E: They existed as the previous method for health host to change the number of cpu cycles it was set to receive this is now handled by the cpu state.

PR #328, opened 2026-08-06
Stress escapees refactor

Gabe-Rose:
# Overview:

## Stress References removed from SGPWorld* and into SGPW_InteractionMechanismSetup.cc
- after_reproduction_sig used to trigger processing of escapees
- emp::vector<StressEscapee> symbiont_stress_escapees and StressEscapee moved to SGPW_InteractionMechanismSettup.cc
- Now uses SymDoBirth

## SymDoBirth modified to take offspring, parent, and world position 
- default mode changed to maintain override
- Relevant functors/lower level functions changed to align with SymDoBirth

## SetupHorizontalTransmissionCompatibilityMode() changed to set one functor
- stress-specific functor removed
- now takes pointers to the (prospective) host and (offspring) symbiont 

## Other:
- Various useless comments removed
- Variables named "baby" renamed to "offspring" where convenient

## Next steps:
- change all .Delete() calls to SendToGraveyard()
- move all sym_offspring SendToGraveyard() calls to SymDoBirth
- ^^^ same with HostDoBirth()?
- Why are we doing the compatibility check with the sym_parent, not the sym_offspring?


source/sgp_mode/SGPWorld.cc
```diff
@@ -197,85 +197,30 @@ emp::WorldPosition SGPWorld::FreeLivingSymDoBirth(
...
 emp::WorldPosition SGPWorld::SymAttemptHorizontalInfection(
-  emp::Ptr<sgp_sym_t> sym_baby_ptr,
+  emp::Ptr<sgp_sym_t> sym_offspring_ptr,
+  emp::Ptr<sgp_sym_t> sym_parent_ptr,
   const emp::WorldPosition& parent_pos
 ) {
-  // TODO - add any signals?
-  const size_t parent_pop_idx = parent_pos.GetPopID();
-  emp::Ptr<Organism> parent = this->GetOrgPtr(parent_pop_idx)->GetSymbionts()[parent_pos.GetIndex() - 1];
   emp_assert(!parent->IsHost());
```
anyaevostinar, line 209: I'm confused where "parent" is being defined now

source/sgp_mode/SGPWorld.h
```diff
@@ -64,7 +65,7 @@ class SGPWorld : public SymWorld {
   // Are host and endosymbiont compatible for horizontal transmission?
   // At the moment, task match based on parent vs current
   // NOTE: arguments can't be const because necessary Host.h/Organism.h functions aren't const
-  using fun_horizontal_transmission_compatibility_check_t = std::function<bool(
+  using fun_horizontal_trans_compatibility_check_t = std::function<bool(
```
anyaevostinar, line 68: This should stay "transmission"

source/sgp_mode/SGPWorld.h
```diff
@@ -356,8 +343,9 @@ class SGPWorld : public SymWorld {
   // Returns a target position for symbiont to horizontally transmit into.
   // Returns std::nullopt if failed to find suitable target position.
   std::function<std::optional<emp::WorldPosition>(
-    size_t,                 /* Parent's host location id in world (pops[0][id])*/
-    emp::Ptr<sgp_sym_t>     /* Pointer to symbiont parent (producing the sym offspring) */
+    emp::Ptr<sgp_sym_t> sym_offspring_ptr,
```
anyaevostinar, line 346: I don't think that you can include parameter names, hence the previous syntanx

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
```diff
@@ -360,10 +361,12 @@ TEST_CASE("FindHostForHorizontalTrans when task matching is required for horizon
 
     emp::Ptr<sgp_host_t> host = emp::NewPtr<sgp_host_t>(&random, &world, &config);
     emp::Ptr<sgp_sym_t> symbiont = emp::NewPtr<sgp_sym_t>(&random, &world, &config);
+    symbiont->GetHardware().GetCPUState().MarkTaskPerformed(8);
```
anyaevostinar, line 364: Why was this necessary to add?

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
```diff
@@ -405,22 +407,24 @@ TEST_CASE("FindHostForHorizontalTrans when task matching is not required for hor
...
```
anyaevostinar, line 410: Ditto

source/sgp_mode/SGPWorld.h
anyaevostinar: Also, why are we now passing the offspring as well as the parent? We should only need whatever symbiont is being used to check for compatibility

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
```diff
@@ -330,7 +331,7 @@ TEST_CASE("FindHostForHorizontalTrans when task matching is not required for hor
       }
       symbiont->GetHardware().GetCPUState().MarkTaskPerformed(8);
       WHEN("Task matching is not required for horizontal transmission") {
-        auto pos_found = world.FindHostForHorizontalTrans(source_id, symbiont);
+        auto pos_found = world.FindHostForHorizontalTrans(offspring, symbiont, symbiont->GetLocation());
```
anyaevostinar, line 334: Got to here, not done with review, but started wondering about why passing offspring and will come back tomorrow

source/sgp_mode/SGPWorld.cc
Gabe-Rose: your right! that is pretty confusing. I wonder why that passed compile

Gabe-Rose: changed to sym_parent_ptr

source/sgp_mode/SGPWorld.h
Gabe-Rose: Removed parameter names. Good question with the offspring; They are not used currently and could easily be removed. we are checking the task profile of the parent with the prospective host, but that doesent quite sit right with me. We are determining the "fate" of the offspring based on the parent, not the symbiont itself. In other words fun_horizontal_transmission_compatibility_check determines the compatibility of the parent and prospective host, but that value is used to determine the compatabilty of the sym and prospective host.

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
Gabe-Rose: I moved it out of the WHEN block below. Initially in my refactor I had:

```
void SGPWorld::SetupFindHostForHorizontalTransmission() {
...
    const bool compatible = fun_horizontal_transmission_compatibility_check(
      *prospective_host_ptr,
      *sym_offspring_ptr
      );
...
}
```
Now:
```
void SGPWorld::SetupFindHostForHorizontalTransmission() {
...
    const bool compatible = fun_horizontal_transmission_compatibility_check(
      *prospective_host_ptr,
      *sym_parent_ptr     //This Line Changed
      );
...
}
```
As mentioned earlier in your comment in SGPWorld.h, initially I thought it would make sense to compare the offspring to the host. Plus, I saw this in Reproduce():
```
offspring_cpu_state.SetParentTasksPerformed(
  hardware.GetCPUState().GetTasksPerformed()
);
offspring_cpu_state.SetParentFirstTaskPerformed(
  hardware.GetCPUState().GetFirstTaskPerformed()
);
```
By putting MarkTaskPerformed() before Reproduce() in the test, I thought it would properly get copied to the offspring, and then the offspring would be able to pass the compatibility check. When I was wrong, I forgot to move this line back and remove the reproduction.

source/sgp_mode/SGPWorld.h
Gabe-Rose: See SGPWorld.test.cc comment as well

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
anyaevostinar: Pretty sure it should be just "symbiont" being passed since offspring has no useful information

source/sgp_mode/SGPWorld.h
anyaevostinar: Let's focus on just refactoring to capture existing functionality and not changing functionality in this PR, since otherwise too many things are changing and it is likely that a bug gets introduced. 

The parent is injecting its offspring into a new host and the offspring hasn't done anything yet, so it can't do its own infecting. The fate of offspring is very often dependent on the parent's behavior ;)

source/test/sgp_mode_test/unit_tests/SGPWorld.test.cc
```diff
@@ -450,23 +454,25 @@ TEST_CASE("FindHostForHorizontalTrans when task matching is required for horizon
 
     emp::Ptr<sgp_host_t> host = emp::NewPtr<sgp_host_t>(&random, &world, &config);
     emp::Ptr<sgp_sym_t> symbiont = emp::NewPtr<sgp_sym_t>(&random, &world, &config);
+    symbiont->GetHardware().GetCPUState().MarkTaskPerformed(8);
     host->AddSymbiont(symbiont);
 
     world.AddOrgAt(host, 0);
-    size_t source_id = symbiont->GetLocation().GetPopID();
+    emp::Ptr<Organism> offspring_org = symbiont->Reproduce();
+    emp::Ptr<sgp_sym_t> offspring = static_cast<sgp_sym_t*>(offspring_org.Raw());
```
anyaevostinar: You'll be able to remove these lines once you make the change to not pass offspring, so please don't forget to do so

anyaevostinar (changes requested):
Some things to change back. Make sure to be careful not to change functionality other than the specific thing that is the target. Also, it's good to clean up small things, but try to resist moving small lines around when they really don't change much since it adds to the complication of reviews and merge conflicts

PR #330, opened 2026-08-11
Bump Empirical pin for hash_namify fix

mmore500:
Opening separately in case there are complications with bumping the Empirical pin.


mmore500:
Looks like there are other complications in bumping the Empirical fix. Happy to take a go at it, but I'll leave the call to primary maintainers.

anyaevostinar:
I'm worried there will be more issues, but the emp_error one looks easy enough. Do you know what emp_error changed to? I'm pretty sure that bumping all the way to the most recent will require some of the fragile tests to be fixed due to the change in RNG

PR #331, opened 2026-08-13
Changed SetReproCount to LineageLength

LaurenC267:


source/default_mode/Host.h
```diff
@@ -280,7 +280,7 @@ class Host: public Organism {
    *
    * Purpose: To set the count of reproductions in this lineage.
    */
-  void SetReproCount(size_t _in) { reproductions = _in; }
+  void LineageLength(size_t _in) { reproductions = _in; }
```
anyaevostinar, line 283: This should be "SetLineageLength" and then the GetReproCount should be changed to "GetLineageLength"

anyaevostinar (changes requested):
Need to put the "Set" back in and also change the "GetReproCount"

